PR #880, opened 2014-11-11
Blender PR

stephanenicolas:
Hi Guice Team,

as discussed on the mailing list, @emmby and I would like to introduce a new annotation processor to Guice to boost its performances. Mike authored RoboGuice and I contribute to it actively too. Our focus is to make guice more efficient so that it can remain an interesting DI option on Android.

We currently use Blender (indeed an extension called RoboBlender) in RG 3 now and it works fine (though we have a few minor issues). According to us, this extension is close to be production ready. Nevertheless, we feel like some points could be enhanced to better match the quality standards of Guice itself. 

So, we submit this PR in order to start talking with the Guice team and see if you guys are interested by this technology. We already got plans to push Guice performance further...

Please note : we couldn't squash all changes as we had a lot of merges to current guice state. Sorry about that.

Roughly speaking, this PR : 
* adds a new module named blender : an annotation processor
* the annotation processor scans all files to compile in order to retrieve injection points and create an annotation database. The injection points will be passed to guice, avoiding a lot of reflection at runtime.
* annotation databases are created as code, this can look a bit ugly, but it would be very difficult to use a binary file in such a way that both Java and Android could use it easily ; Android resource loading being quite specific.
* the annotation database is also used to discard unnecessary bindings (it can sound strange but RoboGuice actually binds a lot of things automatically to make Android development more easy, and we want to keep only the bindings used in an application).
* we also introduced a hierarchy traversal filter to prune the classes to scan for injection points at runtime. Each scan is costly, so we offer to stop scanning when finding some given ancestor classes. For instance, in RG based apps, all classes that use injections will inherit from RG classes, and we can stop scanning when we have processed them, no need to go further up than RG classes (no need to scan their parents, they don't contain injection points..).

You can find much more details in the slides : 
https://speakerdeck.com/stephanenicolas/blender-boosting-guice-with-annotation-processing

Thanks for your work on Guice,
 Stéphane for the RG team.


sameb:
Sorry this took me so long to review.  In a nutshell, this PR request doesn't look ready for inclusion, nor does it look like it's something we'd want to include in Guice anyways.  In terms of merging into Guice, here's the primary issues:
 * The build doesn't compile (see https://travis-ci.org/google/guice/jobs/40686602).
 * The PR contains a lot of unnecessary changes (e.g, rewriting POMs, arbitrary changes in other files unrelated to the purpose of the PR, etc.).
 * Adds complicated APIs to core Guice code without a strong reason.  Specifically,
     - HierarchyTraversalFilter looks to be a pure optimization and can easily introduce accidental bugs.  The cost of the new API and chance of introducing bugs likely doesn't outweigh the gains of the filter.
     - The annotation database finder looks to be a solution solely for a problem that RoboGuice introduces.  e.g, RoboGuice binds a lot of unnecessary stuff, AnnotationDatabaseFinder lets you ignore those bindings.  Instead: just don't bind the things to begin with.   It doesn't make much sense to introduce a lot of code whose purpose basically is: "ignore all these bind calls I'm making".

In terms of a potential fork, you're totally welcome to fork the project and include whatever changes to internals that you'd like.  I'd highly suggest that if you're looking into dependency injection via annotation processing that you instead look into Dagger2.  Guice just isn't designed to work by annotation processing.  Maybe one day it will... and if it does we may look back into this.  But for now, Guice doesn't do it and has no plans to do it.  If we were to do it in the future, we'd likely want to do it without a public API.  That is, there'd be some kind of annotation processor that generates hints for Guice to avoid runtime costs, and it would work without requiring the user to configure it at runtime.

stephanenicolas:
Hi sam,

I have only seen your review today. First thanks for reviewing it.

I perfectly understand that this change is really big. I apologize for the undesired code changes, I was not hoping to get a PR ready for merging, but it was more a way for us to explain the changes we would like to make.  Actually the pom changes are necessary for us, RG team, as we have to constently play with guice artifact of our fork.. Anyway, that was not the point of the PR and I am sure you got that.

Same about compilation. 

About the reason you gave, I don't really agree, and be sure I understand perfectly it doesn't change anything :) 

But, for the sake of completeness, here are a few elements of answer to your comments : 
* HierarchyTraversalFilter can create bugs. For sure, as every change, no ? And you are right, this is about pure optimization of what Guice does. Nothing else mattered to us in this PR. Did you really see that we can get rid of most class scanning via reflection using Blender ? I am surprised you don't mention that gain as it is really huge in terms of performance. And that being considered, I wonder if the change is that big for Guice. Or I would have appreciated any guidance to minimize it.
* you are wrong about the Bindings introduced by RoboGuice. RoboGuice makes it easy to inject a lot of android entities, which are present on the platform. I still strongly believe it would be nice to offer them to users without penalties, and that's what this PR is about. If there was an equivalent with Dagger 2 or Dagger 1 even for what RG does, there would be no penalty to pay to provide more bindings than necessary. It's one of the weakness of Guice and we remedied to it.
* the AnnotationDatabaseFinder cannot be really separated from the HierarchyTraversalFilter, and it's not more related to filtering unused bindings than optimizing class scans. It does the 2 at the same time. I am surprised you consider the AnnotationDatabaseFinder differently. Maybe you have been fast in your review..

I am must say I am a bit sad to see Guice so monolithic and not being able to influence its development. Sure Dagger 2 is gonna be faster, and Dagger 1 is faster. But most of the Guice slowness doesn't come from the way Guice computes values to be injected, it comes from using reflection to find injection points and to assign injected values into fields/methods/constructors. And we continue to think that the gain in speed in daggers DI systems doesn't really balance the added complexity, but that holds only if we can make Guice faster. 

It's unfortunate that we didn't get better feedback from this PR, and even more than it couldn't change your mindset about Guice and its potential on Android and potential to be faster. This lack of feedback and the absence of ability to change Guice development is for us a major reason to switch to another DI, much more than the real performances of Guice. 

FYI, even though this PR was not clean enough to be merged, and again, we said it since the very beginning, Blender and RoboBlender are now in production and drive millions of apps on millions of devices and only one bug has been reported, about assisted injection (we consider it an advanced feature and though we know how to fix, it is postponed). 

I am sorry that my tone is so bitter in my answer and this comment only engages myself, not other RG team contributors. Anyway, I think you are based on the west coast, so we are almost collocated. I would be happy to have a beer and talk about all that if you feel like it.

PR #883, opened 2014-12-05
Avoid duplicate failures if a Singleton fails to create

stevenschlansker:
Current behavior:

If a Singleton cannot be created, it will re-attempt at every injection point.  In `Stage.PRODUCTION`, this is guaranteed wasted work -- the injector creates all Singletons eagerly, and any failure causes Injector creation failure.  In `Stage.DEVELOPMENT`, it's a bit more confusing as we may fail later at runtime, which conceivably allows us to 'recover' from this sort of error.  I would argue this is more a bug than a feature -- it allows multiple partially-created Singleton instances to leak (if you do so from the `Provider` or constructor) and is guaranteed to break later if you ever use `PRODUCTION` mode.

New behavior:

The exception thrown while creating the Singleton is saved and reused at every further try to inject it.  Duplicate exceptions are suppressed from `Errors` to ensure the end user only sees the true cause once.

### DIFF
--- a/core/src/com/google/inject/internal/Errors.java
+++ b/core/src/com/google/inject/internal/Errors.java
@@ -527,6 +527,12 @@ public final class Errors implements Serializable {
     if (root.errors == null) {
       root.errors = Lists.newArrayList();
     }
+    // If this message has already been noted, don't add it again
+    for (Message m : root.errors) {
+      if (m.getCause() == message.getCause() && m.getMessage().equals(message.getMessage())) {
+        return this;
+      }
+    }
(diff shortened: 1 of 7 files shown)


sameb:
Since there seems to be no further activity @stevenschlansker and the CLA isn't signed, I'm going to close this PR.

dluc:
Is there a way to disable the cache of creation failures, regardless of singletons or not? perhaps a setting? Although suboptimal, depending on what they do, ctors can be affected by temporary glitches, and they should be retried.

Issue #888, opened 2014-12-19
Linked bindings have surprising behaviour with private modules / child injectors.


tavianator:
I find the following behaviour surprising:

```java
interface Dependency {
}

interface Interface {
    boolean hasDependency();
}

class Implementation implements Interface {
    @Inject(optional = true) Dependency optional;

    @Override
    public boolean hasDependency() {
        return optional != null;
    }
}

class MyModule extends AbstractModule {
    @Override
    protected void configure() {
        bind(Dependency.class)
                .toInstance(new Dependency() { });

        bind(Interface.class)
                .to(Implementation.class);
    }
}

@Test
public void testOptionalInjection() {
    assertTrue(Guice.createInjector(new MyModule())
            .getInstance(Interface.class)
            .hasDependency());

    // Wha....?
    assertFalse(Guice.createInjector()
            .createChildInjector(new MyModule())
            .getInstance(Interface.class)
            .hasDependency());
}
```

This happens because Guice creates a just-in-time binding for `Implementation`, and just-in-time bindings are added to the root module/injector whenever possible.  But I'd guess that most of the time, that's not what the user intended.

Therefore, I propose a new `Binder` flag, `bindLinkedBindingTargetsLocally()` (or some better name ;), that causes the above to behave as if `bind(Implementation.class);` occurred in the same module as `...to(Implementation.class)` (unless `Implementation` is bound explicitly somewhere else).  Either that or a variant of `requireExplicitBindings()` that applies to linked binding targets too.

Even `OptionalBinder` doesn't solve this problem if the `OptionalBinder` is created in the root injector.

PR #889, opened 2014-12-20
Fix https://github.com/google/guice/issues/888.

tavianator:
When Binder.requireExplicitBindings() is in effect, don't allow any
JIT bindings to be created in parent injectors, even JIT bindings that
are normally exempt such as the targets of linked key bindings.

### DIFF
--- a/core/src/com/google/inject/internal/InjectorImpl.java
+++ b/core/src/com/google/inject/internal/InjectorImpl.java
@@ -101,7 +101,7 @@ final class InjectorImpl implements Injector, Lookups {
     NO_JIT,
     /** allows existing just in time bindings, but does not allow new ones */
     EXISTING_JIT,
-    /** allows existing just in time bindings & allows new ones to be created in the current injector */
+    /** allows existing just in time bindings & allows new ones to be created */
     NEW_OR_EXISTING_JIT,
   }
 
--- a/core/src/com/google/inject/internal/InjectorImpl.java
+++ b/core/src/com/google/inject/internal/InjectorImpl.java
@@ -101,7 +101,7 @@ final class InjectorImpl implements Injector, Lookups {
     NO_JIT,
     /** allows existing just in time bindings, but does not allow new ones */
     EXISTING_JIT,
-    /** allows existing just in time bindings & allows new ones to be created */
+    /** allows existing just in time bindings & allows new ones to be created in the current injector */
     NEW_OR_EXISTING_JIT,
   }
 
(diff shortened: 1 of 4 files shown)


core/src/com/google/inject/internal/InjectorImpl.java
```diff
@@ -101,7 +101,7 @@ public String toString() {
     NO_JIT,
     /** allows existing just in time bindings, but does not allow new ones */
     EXISTING_JIT,
-    /** allows existing just in time bindings & allows new ones to be created */
+    /** allows existing just in time bindings & allows new ones to be created in the current injector */
```
tavianator: This is wrong/misleading; as long as `requireExplicitBindings()` isn't set, `NEW_OR_EXISTING_JIT` still allows them in parent injectors.  Will fix.

sameb:
Sorry for delay in looking at this.  I'm just testing it out over our internal corpus to see if it trips anything up.  Assuming it doesn't, this LGTM.  If it does, may take a bit longer...

sameb:
Looks like this triggered some failures.  Looking more closely at the change, it looks like it's because this changes behavior in an unexpected way.  The rules for the previous behavior were:
  * If parent sets requireExplicitBindings, then all linked bindings will always be local to the injector that created them.  That is, it *already* behaved the way this PR wants.
  * If a parent didn't set requireExplicitBindings but a child did, a linked binding could be promoted to a parent injector.  That is, it didn't behave the way this PR wants.

We should keep the first rule -- if a parent sets requireExplicitBindings, then it's okay for a child to bind(Foo.class).to(FooImpl.class) w/o another explicit bind(FooImpl.class) statement, because the parent knows explicit bindings are required, so it lets the child take it.

We should fix the second rule -- if the parent didn't set requireExplicitBindings, then we should require an explicit binding for FooImpl (because we aren't sure if the parent would want it).

The PR right now fixes the second rule, but in doing so breaks the first one.  If we can leave the first rule intact, that'd make life a lot easier.

tavianator:
> - If parent sets requireExplicitBindings, then all linked bindings will always be local to the injector that created them. That is, it already behaved the way this PR wants.

Really?  Because this test passes against master:

```java
public void testWeirdThingSamBerlinSaid() {
  Injector parent = Guice.createInjector(new AbstractModule() {
    @Override
    protected void configure() {
      binder().requireExplicitBindings();
    }
  });
  Injector child = parent.createChildInjector(new AbstractModule() {
    @Override
    protected void configure() {
      bind(Foo.class).to(FooImpl.class);
    }
  });
  ensureWorks(child, Foo.class);
  ensureFails(child, ALLOW_BINDING, FooImpl.class);
  ensureFails(parent, FAIL_ALL, Foo.class);
  ensureWorks(parent, FooImpl.class);
}
```

tavianator:
> - If parent sets requireExplicitBindings, then all linked bindings will always be local to the injector that created them. That is, it already behaved the way this PR wants.

Ah, I did not know that!  I'll restore that behaviour and add a test for it.

sameb:
Weird.  See the existing testChildInjectorInheritsOption test (before changes in this PR) -- it validates that if parent sets requireExplicitBindings, then the child doing bind(Foo.class).to(FooImpl.class) puts FooImpl in the child, not the parent.

tavianator:
@sameb Yeah sorry that test failed, I just didn't see it at first :P

core/test/com/google/inject/JitBindingsTest.java
```diff
@@ -523,7 +513,109 @@ public void testMembersInjectorsCanBeInjected() {
...
+        }
+      });
+      fail("should have failed");
+    } catch (CreationException expected) {
+      assertContains(expected.getMessage(), jitInParentFailed(ProvByProvider.class));
+      assertEquals(1, expected.getErrorMessages().size());
+    }
+  }
+
+  public void testDeleteThis() {
```
tavianator: Oops ;)

sameb:
Still a few failures across the corpus, I'm going to work with the test/project owners to see what their intention was and try to clean things up.  They may change my mind on the usefulness of changing the behavior, we'll see!

PR #894, opened 2015-01-22
Add support to private module in graph extension

houcheng:
All elements in private module will be draw onto
the result DOT file. Note: the interface that used
in higher level and defined in lower level will be
shown twice in the generated graph.


houcheng:
Hi, I've signed the CLA. Thanks.

PR #897, opened 2015-01-24
Merge internal changes

sameb:


core/src/com/google/inject/internal/MoreTypes.java
```diff
@@ -403,6 +397,16 @@ public boolean isFullySpecified() {
       return stringBuilder.append(">").toString();
     }
 
+    private static void ensureOwnerType(Type ownerType, Type rawType) {
+      if (rawType instanceof Class<?>) {
+        Class rawTypeAsClass = (Class) rawType;
```
tavianator: You can avoid some raw type warnings by using `Class<?>` here too.

cgdecker:
LGTM

sameb:
Thanks @cgdecker.  @tavianator: We'll merge this as-is since it's just a merge of internal changes, and I'll do a subsequent change later with the fixes you pointed out.

Issue #904, opened 2015-02-18
Problem with abstract class and assisted inject after migration to Java 8


davido:
With migration change [1] to Java 8, Gerrit Code Review doesn't start any more: [2]. This diff fixed this: [3]. All is fine with the code on Java 7. Is this expected?

[1] https://gerrit-review.googlesource.com/57852
[2] http://paste.openstack.org/show/176620
[3] https://gerrit-review.googlesource.com/64820

sameb:
Java8 does change a few things, but the error & diff you pasted don't seem to have anything to do with java8.  Can you point to 'com.google.gerrit.server.config.GerritGlobalModule' & 'com.google.gerrit.server.change.Module'?  I'd like to see how the factory is being installed.

davido:
* `com.google.gerrit.server.config.GerritGlobalModule`: [1]
* `com.google.gerrit.server.change.Module`: [2]

[1] https://gerrit.googlesource.com/gerrit/+/master/gerrit-server/src/main/java/com/google/gerrit/server/config/GerritGlobalModule.java
[2] https://gerrit.googlesource.com/gerrit/+/master/gerrit-server/src/main/java/com/google/gerrit/server/change/Module.java

sameb:
This indeed looks like a problem with the way assistedinject scans for factory methods & the way java8 introduces new synthetic/bridge methods.  I have an idea of what's wrong and will write up a testcase + fix.  Thanks for reporting it!

victorr:
Hi,

I have been banging my head against this problem this week. Hopefully the following will aid you in your testing.

Any updates on a fix will be greatly appreciated, as my shop can't wait to upgrade to Java 8.

Kind regards.

```java
package java8;

import com.google.inject.AbstractModule;
import com.google.inject.Guice;
import com.google.inject.Injector;
import com.google.inject.Module;
import com.google.inject.assistedinject.Assisted;
import com.google.inject.assistedinject.AssistedInject;
import com.google.inject.assistedinject.FactoryModuleBuilder;

public class InyeccionAsistida
{

    public interface AbstractClass {}

    public interface AbstractConstructionParam {}

    public interface AbstractFactory <T extends AbstractConstructionParam>
    {
        AbstractClass create( T param );
    }
    
    public static class ConcreteClass implements AbstractClass
    {
        private ConcreteConstructionParam mValue;

        @AssistedInject
        ConcreteClass( @Assisted ConcreteConstructionParam value )
        {
            mValue = value;
        }
        
[...]

PR #905, opened 2015-02-20
Merge internal changes

sameb:


sameb:
FYI, cgruber has a fix for the dagger-interop, and I'm working on a fix for the rest of #904... but it's a PITA to get right.   Have some ideas, will try them out this weekend or next week.  Will wait on this PR till I get that & the dagger-interop fix in.

sameb:
Ok this should be good now @cgdecker  -- the 'compareBuilds' script is failing because of some problems with "dagger-adapter" having a dash in its name (see the bottom of https://api.travis-ci.org/jobs/52063347/log.txt?deansi=true), but it doesn't seem to cause the build to error out & it's not fatal anyway.  We can fix later.

cgdecker:
LGTM

PR #920, opened 2015-04-28
Merge internal changes.

cgdecker:
`@since` and `package-info.java` changes.

### DIFF
--- a/core/src/com/google/inject/internal/package-info.java
+++ b/core/src/com/google/inject/internal/package-info.java
@@ -17,5 +17,4 @@
 /**
  * <i>Guice</i> (sounds like like "juice")
  */
-
 package com.google.inject.internal;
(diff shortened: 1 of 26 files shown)


core/src/com/google/inject/internal/package-info.java
```diff
@@ -17,5 +17,4 @@
 /**
  * <i>Guice</i> (sounds like like "juice")
  */
-
-package com.google.inject.internal;
```
sameb: This diff is hard to read.  The diff *looks* like it's deleting the package line.  But the file it's merging from has the package line (and so did the internal CL this was created from).  The package line is staying, right?

cgdecker: It's staying. You can look at the full file at this revision here: https://github.com/google/guice/blob/merge-internal-changes/core/src/com/google/inject/internal/package-info.java

sameb:
LGTM

PR #921, opened 2015-04-29
Added option to configure PersistFilter to assume on initialization that persistence service is already started

lextiz:
Added option to configure PersistFilter not to start persistence service on initialization. Such option is needed because filter is initialized at certain stage of servlet container lifecycle, and connection to the database might be needed before this stage, so it is logical to have an option to start the persistence service manually.


lextiz:
I signed it!

PR #922, opened 2015-04-29
Added option to configure PersistFilter to assume persistence service is initialized

lextiz:
Added option to configure PersistFilter not to start persistence service on initialization. Such option is needed because filter is initialized at certain stage of servlet container lifecycle, and connection to the database might be needed before this stage, so it is logical to have an option to start the persistence service manually.


sameb:
Sorry for the delay in responding to this.  At minimum it needs a text.  Even better would be if this behavior could be controlled through something in the persist module setup instead of a hidden string in the config.

lextiz:
Thanks for the answer, I will add a test. As far as I understand from servlet specification FilterConfig is standard way for passing configuration to servlet filters, please correct me if I am not right.

lextiz:
I will resubmit the pull request

Issue #924, opened 2015-05-06
Grapher generates dot files with syntax errors in v3.0


laurenthuang:
There is same problem as https://github.com/google/guice/issues/755 in version 3.0 of guice-grapher.

The Graphviz I using is version 2.38. Below is the output:
working on D:/codebase/lh.test.automation/org.eclipse.xtext.example.domainmodel.tests/GuiceBinding.dot
Error: <gvedit>: syntax error in line 3 near ','

sameb:
Please use the latest version where this is already fixed.  We don't do maintenance releases on old versions, sorry.

Issue #755, opened 2014-07-07
Grapher generates dot files with syntax errors


gissuebot:
_From [stantonk](https://code.google.com/u/103514229777971627206/) on June 19, 2013 15:54:36_

Description of the issue: Grapher generates dot files with syntax errors:

Warning: graph.dot:3: syntax error in line 3 near ',' Steps to reproduce: 1. Follow the example https://code.google.com/p/google-guice/wiki/Grapher 2. Run the code
3. Try to generate a diagram using dot, e.g.

&nbsp;&nbsp;dot -Tsvg graph.dot &gt; graph.svg

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=755_

Issue #663, opened 2014-07-07
Grapher is producing corrupt graphs


gissuebot:
_From [glenviewjeff](https://code.google.com/u/110087215095127878251/) on October 22, 2011 13:15:56_

My graphs look terrible (there are lines without parent connections.) I assume it's related to these multiple "unknown tag" errors with even simple projects when running dot -Tpng test.dot &gt;test.png:

Error - pango_parse_markup: Unknown tag 'java.lang.Object' on line 1 char 196 

It looks as if dot is parsing the &amp;lt &amp;gt as literal less than/greater than symbols. For instance in the .dot file, the java.lang.Object looks like this:

...&amp;lt;java.lang.Object&amp;gt...

**Attachment:** [gist](https://gist.github.com/e971e8adad7111180283)
&nbsp;&nbsp;&nbsp;_[test.dot](https://gist.githubusercontent.com/gissuebot/e971e8adad7111180283/raw/d035280bedd5a86c9df4691d12352231db7ed4f7/test.dot)_
**Binary attachments:** [test.png](http://code.google.com/p/google-guice/issues/detail?id=663)

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=663_

PR #925, opened 2015-05-06
3.x maintenance

laurenthuang:
There are same problems #755 #663 in version 3.0 of guice-grapher( I opened a ticket #924), and made a patch according to version 4.0.


sameb:
Please use 4.0, where this is already fixed.  We don't do maintenance on old releases, sorry.

PR #929, opened 2015-05-20
Merge internal changes

sameb:
Reduce the amount of stack trace spam in error message.  If all the errors have
the same cause, don't list them individually and instead use as the Throwable's
cause.  If some (but not all) errors are duplicates, elide the duplicate stack
traces and point to the error # it's the same as.

"same" here is defined as the same exception class, same stack trace, and same causes (with causes using the same 'same' definition).
-------------
Created by MOE: http://code.google.com/p/moe-java
MOE_MIGRATED_REVID=92389485

### DIFF
--- a/core/src/com/google/inject/internal/Errors.java
+++ b/core/src/com/google/inject/internal/Errors.java
@@ -16,9 +16,13 @@
 
 package com.google.inject.internal;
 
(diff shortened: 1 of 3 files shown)


cgdecker:
LGTM pending travis build success.

Issue #940, opened 2015-07-03
Message.hashCode and Message.equals are Inconsistent 


tlantz:
I was recently working on something where because of some exception badness, a process was going into an infinite dependency resolution loop. The root cause was definitely in the code using Guice. However, before getting to itI noticed something strange while looking at stack dumps of the process. Nearly every one was in the `ImmutableSet` implementation in Guice.

Debugging a bit further I found that an internally managed set of `Message` objects containing exception messages being thrown by the calling code was taking a very long time to find a slot in which to insert a new `Message`. Unfortunately while every `Message.message` was exactly the same, the `sources` and `throwable` were unique objects so the messages did not have value equality. Because `Message.hashCode` is based on `Message.message` only and doesn't take the other two fields used in `Message.equals` into account, each insertion needed to be rehashed a number of times equal to the number of `Message`s with the same text inserted previously.

The bigger problem was of course the buggy code causing the infinite dependency resolution loop. However, the inconsistent `equals`/`hashCode` implementations cause things to perform in unexpected ways and were distracting to the debugging process.

* Stack dumps were almost always ending up in the middle of collection code which made it harder to figure out what the real cause of the problem was.
* Memory was being consumed very slowly because of the quadratic performance relative to the number of exceptions being thrown. I think that when something goes so terribly wrong, a quick death by memory consumption would be preferred.

sameb:
Note that the problem is not that the two methods are inconsistent. It's
perfectly acceptable for hashCode to be computed with fewer fields than
equals. So long as every equal object has the same hash code, it's OK for
some unequal objects to also have the same hash code. The result is that
fewer hash buckets address used and there'll be more O(n) lookups instead
of O(1) lookups. Memory consumption will not be meaningfully different --
only CPU consumption.

Is it actually common for two errors to have the same message but different
sources or causes?

tlantz:
@sameb: you're correct. I might have used the wrong wording with "inconsistent". I agree `Message` equality is _functionality correct_ as long as `equals` is based off of at least the data used in `hashCode` (and they are).

As you've stated, this is a performance problem (and I'd argue, more importantly, a "practical debugability" problem): `Message`s with the same `message` are colliding but are not equal resulting in O(n) set insertion per `Message`. 

I also agree that it won't change absolute memory consumption, but it does change memory consumption over time (practically) in the sense that when you're accumulating many `Message`s in a set, and those insertions are O(n), your code is spending O(n^2) inserting `Message`s. This has two effects:

[...]

PR #941, opened 2015-07-03
Make Message.hashCode Consistent with Message.equals

tlantz:
Fixes https://github.com/google/guice/issues/940

This commit makes `Message.hashCode` and `Message.equals` computed off of the same fields.

### DIFF
--- a/core/src/com/google/inject/spi/Message.java
+++ b/core/src/com/google/inject/spi/Message.java
@@ -16,6 +16,8 @@
 
 package com.google.inject.spi;
 
+import static com.google.common.base.Preconditions.checkNotNull;
+
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Binder;
@@ -25,8 +27,6 @@ import com.google.inject.internal.util.SourceProvider;
 import java.io.ObjectStreamException;
 import java.io.Serializable;
 import java.util.List;
-
-import static com.google.common.base.Preconditions.checkNotNull;
 
 /**
  * An error message and the context in which it occured. Messages are usually created internally by
--- a/core/src/com/google/inject/spi/Message.java
+++ b/core/src/com/google/inject/spi/Message.java
@@ -16,8 +16,6 @@
 
 package com.google.inject.spi;
 
-import static com.google.common.base.Preconditions.checkNotNull;
-
 import com.google.common.base.Objects;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Binder;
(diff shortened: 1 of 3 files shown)


core/src/com/google/inject/spi/Message.java
```diff
@@ -28,6 +26,8 @@
 import java.io.Serializable;
 import java.util.List;
 
+import static com.google.common.base.Preconditions.checkNotNull;
```
tavianator: Why'd you move this?

tavianator:
Have you measured the performance effect of this?

core/src/com/google/inject/spi/Message.java
tlantz: Inadvertent (IntelliJ defaults). I will fix it.

tlantz:
@tavianator - yes, for the case I was describing above. Inserting single `Message` objects is no longer significant as the number of messages grows. I can post some repro code and results.

The actual implementation of `Message.hashCode` will be slower but I'm not sure that's a very helpful measure. Maybe I can provide the breakeven point in number of `Messages` when the slow insertion overtakes the time spent in individual `hashCode` calls?

Are there other specific benchmark cases you think it would make sense for me to measure? Happy to post results for those as well.

Will try to put something together by the weekend (I don't have much time for it during the work week).

tlantz:
I tested out two things related to performance. For the sake of being thorough I just timed the `hashCode` implementation and as expected the new one is slower. On my machine (1.3GHz Intel i5) 100 million calls of the current implementation is 237 millis. With the new implementation and a source list size of 10 (a bit arbitrary, but didn't seem unreasonable after logging in a real application) 100 million calls is 5320 millis. That's a large percentage increase but still, in nanosecond territory per call. So I'd argue that in any practical application it should not matter.

The second thing I did was time a simpler version of what was happening in the production codebase where I saw the original issue: each time a new `Message` was created with a unique `Throwable` it was merged with all prior `Message`s in an immutable set. I ran the test over a range (log scale) of set sizes. Somewhere between 128 and 256 `Message`s in the set is where it breaks even. 

```
  MessageCount OldHashCodeMillis NewHashCodeMillis
1           32                 3                 5
2           64                 9                11
3          128                15                21
4          256                77                33
5          512               176               103
6         1024              1287               130
7         2048              7995               372
```

The red line below is the timings with the old implementation, the blue line with the new. 

image

The timing code is pretty trivial: just a run to get rid of JIT overhead and then a loop to measure the hash code time or to build sets of a certain size.. 

These message counts are large for any normal use case, but going back to the original issue: when stuck in an infinite cycle of exceptions during dependency resolution in a production situation, this made understanding the current state of an application difficult since the application was spending the majority of its time inserting `Message` objects into a hash set.

Happy to look at specific cases where this change might be considered risky.

core/src/com/google/inject/spi/Message.java
tlantz: Fixed in d20bf5b

Issue #937, opened 2015-06-27
Java 8: Static interface methods considered with FactoryModuleBuilder


kchung-rmn:
Consider the following assisted inject factory interface:

    public interface Factory {
        Impl build();
        
        static Factory getDefault() {
           // ...
        }
    }

This is, of course, permissible in Java 8 as you can associate static methods with interfaces.  However, when you use `FactoryModuleBuilder` on it, as so:

    install(new FactoryModuleBuilder().build(Factory.class));

There is an error along the lines of the following:

    Factory is an interface, not a concrete class. Unable to create AssistedInject factory.

This does not seem to be intuitive, because `Factory` is supposed to be an interface. Debugging this code reveals that this is in reference to the static method, which should, by all means, not be considered because of its static-ness.  Indeed, `FactoryProvider2`'s constructor does not appear to skip over static methods, as expected.

Of course, a workaround exists simply by moving the static method away from the factory interface, but this seems like a bug in the functionality of `FactoryProvider2` specifically.

sameb:
Yup, this seems wrong.  I'll happily merge a PR that skips static methods.

PR #942, opened 2015-07-06
assistedinject: Skip static methods on interfaces.

tavianator:
This fixes https://github.com/google/guice/issues/937.


sameb:
Thanks Tavian!

PR #952, opened 2015-09-11
Close stream after reading line numbers

zlosch:


zlosch:
I signed it!

sameb:
Could you revert the "add missing null check" commit, since I separately merged that one?  Thanks!

tavianator:
Well don't revert it, that would revert it in master too!  Just rebase this branch.

tavianator:
@sameb Actually you can probably just merge this as-is, it's the exact same commit hash so it's not going to get double-applied or anything.

core/src/com/google/inject/internal/util/LineNumbers.java
```diff
@@ -62,7 +62,14 @@ public LineNumbers(Class type) throws IOException {
     if (!type.isArray()) {
       InputStream in = type.getResourceAsStream("/" + type.getName().replace('.', '/') + ".class");
       if (in != null) {
-        new ClassReader(in).accept(new LineNumberReader(), ClassReader.SKIP_FRAMES);
+        try {
+          new ClassReader(in).accept(new LineNumberReader(), ClassReader.SKIP_FRAMES);
+        } finally {
+          try {
+            in.close();
```
tavianator: You can do `Closeables.closeQuietly(in)` to avoid the ugly nested try block.

sameb:
Thanks @tavianator!

@zlosch  -- if you can change to use closeQuietly and rebase, I'll merge in.  Thanks!

zlosch:
Done. Changed to Closeables.close(in,true) because closeQuietly is not available in Guava 16.

tavianator:
@zlosch Since https://github.com/google/guice/commit/effca1d657cb76f3c616643b31cc729efe4b7927, Guice is using Guava 18.  But now that I look more closely at the docs, `Closeables` is `@Beta` so maybe Guice shouldn't be using it.  @sameb thoughts?

sameb:
Indeed, Closeeables.close is beta.  So we can't use it, sorry about that @zlosch.  If you're able to switch it back to what you had before, that'd be better.  Thanks, and sorry!

zlosch:
Change was already done in another commit.

Issue #788, opened 2014-07-07
jpa: Object-methods like finalize become transactional, if the Type is annoptated @Transactional


gissuebot:
_From [google-auth-21ov9@codesmell.de](https://code.google.com/u/113306752357053708143/) on December 04, 2013 09:06:25_

Description of the issue: Steps to reproduce: 1. Annotate a given Type as `@`Transaction (not on method, but on type level)
2. Once the object gets finalized, the Interceptor kicks in

at com.google.inject.persist.jpa.JpaPersistService.begin(JpaPersistService.java:70)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.persist.jpa.JpaLocalTxnInterceptor.invoke(JpaLocalTxnInterceptor.java:49)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.internal.InterceptorStackCallback$InterceptedMethodInvocation.proceed(InterceptorStackCallback.java:72)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.internal.InterceptorStackCallback.intercept(InterceptorStackCallback.java:52)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at xxx.xxx.xxx.xxx.xxx.xxx$$EnhancerByGuice$$5293155f.finalize(&lt;generated&gt;)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at java.lang.ref.Finalizer.invokeFinalizeMethod(Native Method)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at java.lang.ref.Finalizer.runFinalizer(Finalizer.java:101)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at java.lang.ref.Finalizer.access$100(Finalizer.java:32)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at java.lang.ref.Finalizer$FinalizerThread.run(Finalizer.java:190)

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=788_

PR #958, opened 2015-10-25
#788 jpa: Object-methods like finalize become transactional, if the Type is annotated Transactional

andresviedma:
I had the problem described in the issue too. The thing is that when a class is annotated as Transactional, the methods of its superclasses shouldn't be considered transactional. The most serious problem comes with Object methods, and especially finalize. I've fixed that, checking that the method class has the annotation, instead of the class of the called object.

I've also created a "lenient" mode, so that if anyone was using the interceptor in some strange way with no real Transactional annotational attached, it keeps working. But the normal behaviour configured in the module is to be not lenient (strict).


andresviedma:
I signed it!

andresviedma:
Hi! Is anyone reviewing the pull requests? This PR is already 6 months old...

Issue #731, opened 2014-07-07
Add @UnitOfWork annotation


gissuebot:
_From [spootsy.ootsy](https://code.google.com/u/103989710404024834643/) on October 05, 2012 00:28:48_

Something that seems to be missing from guice-persist is a `@`UnitOfWork equivalent to the `@`Transactional annotation. One that will automatically start/stop a UnitOfWork as required, but not begin/commit a transaction.

Our use case: we have a bunch of classes that are used for database access. We use these in command-line applications, both during initialisation (in provider methods) and at run-time.

If a method exists for altering data in the database, we can happily apply `@`Transactional to it. It can now be safely called whether you're in an active UnitOfWork or not (because `@`Transactional will only start/stop a UnitOfWork if required).

For methods that only read data from the database, we don't want to start/commit transactions needlessly, so the `@`Transactional tag is a no-go. But - we have no means of guaranteeing that a method is encapsulated by a UnitOfWork automatically... there's no way to check if a UnitOfWork is already running, so there's no way to start/stop one only as required.

Would be a boon to have this functionality. Alternatively, an "isRunning()" method on UnitOfWork would make it possible for us (and others) to create our own automatic UnitOfWork controllers.

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=731_

PR #959, opened 2015-10-25
#731 Add @RequiresUnitOfWork annotation

andresviedma:
Implementation of a @RequiresUnitOfWork annotation equivalent to the @Transactional annotation. One that will automatically start/stop a UnitOfWork as required, but not begin/commit a transaction. See issue #731 for more info. I think this annotation can be interesting, I have also missed it often.

The name of the annotation is important as it's public, and so can be changed. Please check if you like it and change it accordingly.

This pull request also includes a refactoring of JpaLocalTxnInterceptor class, extracting the management of the unit of work to a new utility UnitOfWorkHandler class, which can also be handy to start / end units of work by code (instead of with annotations). Finally, it also solves issue #753 (JpaLocalTxnInterceptor is inconsistent. It uses both UnitOfWork and JpaPersistService), by using the same UnitOfWork injected object to both begin and finish it.


andresviedma:
Hi! Is anyone reviewing the pull requests? This PR is already 6 months old...

tuxbear:
Guice is dead?

PR #979, opened 2016-01-15
JUnit 4 Guice Persist functionality. 

jccarrillo:
Provides a custom runner to assist with JUnit 4 test cases that require usage of guice persist.
JUnit 4 Runner that Guicifies the test, installs and manages guice persist.
Adds following behavior:
- Guicifies the test.
- Initializes the JpaPersistModule.
- Initializes any other provided Modules

Usage:

```
@RunWith(PersistJUnit4ClassRunner.class)
@PersistJUnit4Context(
    jpaUnit = "testUnit", 
    jpaPropertiesClass = SampleJpaProperties.class, 
    modules = {SampleModule.class})
public class SomeJUnit4Test {

  @Inject
  private SomeInjection injected;

  @Test
  public void assertInjection() {
    Assert.assertNotNull(this.injected);
  }
}
```


extensions/persist/pom.xml
```diff
@@ -27,6 +27,12 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.11</version>
+      <scope>compile</scope>
```
tavianator: This makes JUnit a runtime dependency of `guice-persist`, which is probably not what most people want.

extensions/persist/pom.xml
```diff
@@ -27,6 +27,12 @@
       <scope>provided</scope>
     </dependency>
     <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <version>4.11</version>
+      <optional>true</optional>
```
jccarrillo: Added optional true to prevent inheritance of junit.

jccarrillo:
Hi, need assistance with the build. It's not responding.

sameb:
Thanks!  I don't think we want this in the actual extension though.  Seems wrong to have test-only code included, even if the JUnit dep itself is optional. We have a separate testlib extension. Could it fit there instead?

jccarrillo:
Certainly. Would a persist-junit module be overkill?

sameb:
Yeah, I think so. Bundling into the existing testlib should be fine.

jccarrillo:
Moved as @sameb suggested.

sameb:
Hello!  Thank you for this PR. Things have changed a bit since this PR was submitted, and also the CLA isn't signed... so I'm going to close this PR.  Please re-open if you'd like to pursue it, thanks!

PR #982, opened 2016-01-16
ignore everything in extensions' build dir

jccarrillo:
When running the 'ant' build a bunch of files get created within the extensions' build directories.

These are now being ignored.


sameb:
Thanks!

PR #987, opened 2016-01-27
Fix Checkstyle violation: forbidding making public Logger

yaziza:
Return Logger out of the class is a very bad practice as it produce logs that are hard to investigate as logging class does not contains that code and search should be done in other classes or in hierarchy (if filed is public or accessible by other protected or package)

### DIFF
--- a/core/src/com/google/inject/internal/BytecodeGen.java
+++ b/core/src/com/google/inject/internal/BytecodeGen.java
@@ -63,7 +63,7 @@ import java.util.logging.Logger;
  */
 public final class BytecodeGen {
 
-  static final Logger logger = Logger.getLogger(BytecodeGen.class.getName());
+  private static final Logger logger = Logger.getLogger(BytecodeGen.class.getName());
 
   static final ClassLoader GUICE_CLASS_LOADER = canonicalize(BytecodeGen.class.getClassLoader());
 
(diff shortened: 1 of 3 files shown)


sameb:
Does Guice even have a checkstyle config that this is violating?

yaziza:
Hi @sameb, sorry for the confusion.

I am a Checkstyle contributor, i sent this PR to get some feedback about a new check we have that forbid public Logger:

https://github.com/sevntu-checkstyle/sevntu.checkstyle/issues/286

https://github.com/sevntu-checkstyle/sevntu.checkstyle/pull/426

Guice was one of the project i tested and wanted to share the results with you.

Thank you !

core/src/com/google/inject/internal/BytecodeGen.java
```diff
@@ -63,7 +63,7 @@
  */
 public final class BytecodeGen {
 
-  static final Logger logger = Logger.getLogger(BytecodeGen.class.getName());
+  private static final Logger logger = Logger.getLogger(BytecodeGen.class.getName());
```
tavianator: This `Logger` is used from within an (anonymous) inner class [here](https://github.com/yaziza/guice/blob/checkstyle/core/src/com/google/inject/internal/BytecodeGen.java#L137).  Making it private causes the field to be accessed indirectly through a synthetic method, which I'm assuming is why it was package-private in the first place.  Since it's not like it was `public` before, IMO this change actually reduces the code quality.

yaziza:
Hi @tavianator, thank you for your feedback !

PR #989, opened 2016-02-07
A more usable BOM and a new Guice Platform BOM

jccarrillo:
Refactored pom files to provide a more usable BOM and a new Guice Platform BOM to ease Guice Integration.

### DIFF
--- a/bom/README.md
+++ b/bom/README.md
@@ -1,0 +1,35 @@
+Guice Bill of Materials (BOM)
+============================
+This BOM contains all the Guice modules in a single dependency management.
+
+Use as follows:
+
+```
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+    <modelVersion>4.0.0</modelVersion>
+    
+    <groupId>com.myproject</groupId>
+    <artifactId>myartifact</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>com.google.inject</groupId>
+                <artifactId>guice-build</artifactId>
+                <version>4.0.1-SNAPSHOT</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    ...
+</project>
+```
+
+For more information about [Dependency Management](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html)
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -5,7 +5,7 @@
 
   <parent>
     <groupId>com.google.inject</groupId>
-    <artifactId>guice-parent</artifactId>
+    <artifactId>guice-build</artifactId>
     <version>4.0.1-SNAPSHOT</version>
   </parent>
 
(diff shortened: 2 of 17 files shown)


sameb:
Thanks for the patch!

@tbroyer, you initially added the BOM, and @mcculls, you know the most about Maven... WDYT of this?

jccarrillo:
With these changes, users can use the Guice BOM as a parent and override individual dependencies. For instance, if you wish to downgrade guice persist, you can just add the version property and that is it. Master doesn't allow you to do that, a user would need to add a dependency management entry.

Also, I thought of the "platform" as a way to easily inherit all the Guice dependency management. For instance, Guice uses google dagger and the version being used is 2.0, a user inheriting from the the platform will not need to specify the version of dagger BUT if the user wants to upgrade it to 2.0.2 (current), only the version property is needed.

Examples are in the README.md of each module.

Guice BOM: https://github.com/jccarrillo/guice/tree/bom/bom

Guice Platform BOM: https://github.com/jccarrillo/guice/tree/bom/platform

Thanks for looking at this.

tbroyer:
>  Master doesn't allow you to do that, a user would need to add a dependency management entry.

What's wrong with that?

Actually, I don't think one should use a BOM as a parent POM.
If you want to override a dependency-management entry from a BOM, either add it as a dependency with a version (and/or scope), or as a dependency-management.
Using properties and a "BOM as parent POM" feels just wrong to me; even more so that its only "added value" is to save you a few lines (override property value vs. override dependency-management entry), and it comes at the expense of maintainability (one day, a developer will come, see that property, grep for it, find nothing, and simply delete it; if you want a dependency-management entry, use a dependency-management entry; verbosity should not be an excuse).

I'm OK with introducing a Guice BOM separate from guice-parent, but only as a way to separate build-related dependencies (e.g. JUnit) and other transitive dependencies (Dagger, Spring, etc.) from Guice artifacts.

I don't think the “platform” BOM is useful (see above wrt usage of properties).

Similarly, Guava should IMO be left as a transitive dependency and its version not fixed in the BOM.

jccarrillo:
It's not about being right or wrong. It's about giving users "options".
A user can import a bom in the dependency management and apply those overrides in the dm as always.

Being come from a spring background, the usage of spring's platform as a parent is beneficial for me and I'm sure plenty of other developers.

Again, this patch is not a must use as a parent, it's about providing the user the option of using it as a parent, either the Guice BOM or the Guice Platform BOM.

See springs platform documentation as reference: http://docs.spring.io/platform/docs/current/reference/html/getting-started-using-spring-io-platform.html

Regards,

-JC

bom/README.md
```diff
@@ -0,0 +1,73 @@
+Guice Bill of Materials (BOM)
+============================
+This BOM contains all the Guice modules in a single dependency management.
+
+There are 2 ways of utilizing the BOM.
+
+1. Using it as the parent of your application's pom.
```
mcculls: Personally I'd discourage using the BOM as a parent pom for your application because of the way that pom inheritance works. I also suspect not many people would actually use the BOM in this way, most people would already be using an OSS or corporate pom (or want more control over the pom) in which case it's much easier to [import](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Importing_Dependencies) the BOM into your application pom.

bom/pom.xml
```diff
@@ -15,12 +15,30 @@
...
+    <guice-jmx.version>4.0.1-SNAPSHOT</guice-jmx.version>
+    <guice-jndi.version>4.0.1-SNAPSHOT</guice-jndi.version>
+    <guice-multibindings.version>4.0.1-SNAPSHOT</guice-multibindings.version>
+    <guice-persist.version>4.0.1-SNAPSHOT</guice-persist.version>
+    <guice-servlet.version>4.0.1-SNAPSHOT</guice-servlet.version>
+    <guice-spring.version>4.0.1-SNAPSHOT</guice-spring.version>
+    <guice-struts2.version>4.0.1-SNAPSHOT</guice-struts2.version>
+    <guice-testlib.version>4.0.1-SNAPSHOT</guice-testlib.version>
+    <guice-throwingproviders.version>4.0.1-SNAPSHOT</guice-throwingproviders.version>
+    <guava.version>18.0</guava.version>
```
mcculls: I find this use of version properties odd, normally version properties are used sparingly - for example when you have several related dependencies and want to apply the same version across them. Using them to support overriding any dependency version in child projects is IMHO an anti-pattern as it makes the build less deterministic.

This change also seems specific to using the BOM as a parent pom, which I'd discourage for the reasons above. It's much simpler to import the BOM and then override specific dependency versions using `<dependencyManagement>`.

bom/pom.xml
```diff
@@ -29,70 +47,64 @@
         <classifier>no_aop</classifier>
       </dependency>
       <dependency>
-        <groupId>com.google.inject</groupId>
-        <artifactId>guice</artifactId>
-        <version>${project.version}</version>
-        <classifier>classes</classifier>
```
mcculls: This entry should be kept in the BOM as it represents the original non-jarjar'd classes for people who want to consume Guice and use it with a different version of CGLIB/ASM.

extensions/assistedinject/pom.xml
```diff
@@ -13,4 +13,21 @@
 
   <name>Google Guice - Extensions - AssistedInject</name>
 
+  <dependencies>
+    <dependency>
+      <groupId>com.google.inject</groupId>
+      <artifactId>guice</artifactId>
```
mcculls: All extensions depend on Guice core, so it looks odd to have to repeat this in each pom rather than simply declare it once in the extensions parent pom as it was before.

parent/pom.xml
```diff
@@ -0,0 +1,170 @@
...
+    <dependencies>
+      <dependency>
+        <groupId>com.google.inject</groupId>
+        <artifactId>guice</artifactId>
+        <classifier>tests</classifier>
+        <version>${guice.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.inject</groupId>
+        <artifactId>guice-bom</artifactId>
```
mcculls: Seems odd to import the BOM into the build's parent pom, when it's conceptually an output of the build.

platform/README.md
```diff
@@ -0,0 +1,73 @@
+Guice Platform Bill of Materials (BOM)
+============================
+This BOM is an extension of of the Guice BOM. It appends all dependencies needed to run Guice and its extensions.
```
mcculls: I don't quite get the difference between the BOM and platform BOM, I think it might confuse people having two BOMs.

mcculls:
@sameb I agree with the points that @tbroyer already made (I also added some more comments inline)

extensions/assistedinject/pom.xml
jccarrillo: adding this to the parent pom as for all to inherit this dependency.

platform/README.md
jccarrillo: The Guice BOM represents Guice Components. The Guice Platform represents the Guice Components and all those dependencies that make Guice work.

bom/pom.xml
jccarrillo: See some examples:
1. https://github.com/spring-io/platform/blob/master/platform-bom/pom.xml
2. https://github.com/spotify/apollo/blob/master/apollo-bom/pom.xml
3. https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-bom/pom.xml

A user is not loosing the ability to used the BOM by overriding dependency versions using `<dependencyManagement>`.

bom/pom.xml
jccarrillo: I am putting it back.

bom/pom.xml
mcculls: Only 1. shows this pattern... 2. follows the practice of using a version property when you have related dependencies you want to tie together (like cucumber-java + cucumber-junit) and 3. doesn't appear to use version properties at all.

platform/README.md
mcculls: The only difference I can see is the Guice Platform adds further version properties for transitive dependencies of components already in the Guice BOM. So this extra pom is only useful if you want to override these transitive dependency versions using properties - but really you should be using `<dependencyManagement>` to do this because a) it's the standard way to override versions and b) it works for all dependencies, whereas the version properties approach only works for projects/BOMs which follow the same property pattern

bom/pom.xml
jccarrillo: I just tried `mvn release:update-versions` to make sure those versions get updated and they did not. Bringing {$project.version} back.

bom/pom.xml
```diff
@@ -15,84 +15,102 @@
...
+    <guice-multibindings.version>${project.version}</guice-multibindings.version>
+    <guice-persist.version>${project.version}</guice-persist.version>
+    <guice-servlet.version>${project.version}</guice-servlet.version>
+    <guice-spring.version>${project.version}</guice-spring.version>
+    <guice-struts2.version>${project.version}</guice-struts2.version>
+    <guice-testlib.version>${project.version}</guice-testlib.version>
+    <guice-throwingproviders.version>${project.version}</guice-throwingproviders.version>
+    <guava.version>18.0</guava.version>
+  </properties>
+
```
jccarrillo: Removed the hardcoded versions as they didn't update during `mvn release:update-versions` but kept the properties. I believe this gets the best of both worlds.

Thoughts?

extensions/assistedinject/pom.xml
jccarrillo: Moved it to Guice BOM

jccarrillo:
I have addressed all concerns, if there are any others, please let me know. Thanks for having looked at this patch!

mcculls:
@jccarrillo I still have concerns with this patch:
* I really don't think we should be promoting use of the BOM as a parent POM, so any changes relating to this should be removed from the patch
* The overuse of version properties should be removed or scaled back, since most of those are specific to using the BOM as a parent POM
* The platform BOM doesn't bring much value and should be dropped to avoid confusion

Since the original goal was to improve the BOM, perhaps it would be better to start with with changes focused on that part of the build, such as disconnecting it from guice-parent, and avoid changing other non-BOM parts of the build?

jccarrillo:
@mcculls,

- Promotion of BOM as a parent has been removed and changes related to that has been removed as well.
- Version properties have been reduced to two, guice and guice-extentions.
- The platform BOM has been removed.

Please let me know if there is anything else needed.

sameb:
Sorry that this was left for so long. Things have likely shifted a lot things it was sent as far as the poms go, so I'm going to close this.  If you'd like to pursue it, please re-open and we can go from there.

Issue #739, opened 2014-07-07
Guice-persist: Injection of EntityManager outside a UnitOfWork can lead to never closed EntityManagers


gissuebot:
_From [head@anthologique.net](https://code.google.com/u/108548559019087008997/) on February 02, 2013 05:40:00_

Form the commit rb7a02b02d81c (Jul 7, 2011), in JpaPersistService.java which is a Provider&lt;EntityManager&gt;, in the get() method, we can see that if we don't already have an EntityManager stored in the thread local variable, we call the begin() method. This method will create a new EntityManager, and store it in the thread local variable.

So, if we follow the documentation in the wiki (page "Using JPA", section "Using the EntityManager inside transactions"), for each thread, for every other action needing an EntityManager, the same instance will always be returned, and this EntityManager will never be closed.

This can lead to broken threads, as if for any reason the EntityManager can no longer be used (connection closed, etc.), there is no way to replace it.

Note that this issue doesn't happen if the injection of EntityManager is done inside an opened UnitOfWork or inside a method annotated with `@`Transactional. In these cases, the EntityManager is already created, and will be destroyed at the end of the transaction, or on UnitOfWork close.

While taking a look at the get() method of JpaPersistService, we can also see that there is a strange precondition test. As said, if the EntityManager doesn't exist, it's created. Juste after that, we check if we have an EntityManager. If we don't have any, we throw this error message: "Requested EntityManager outside work unit.". Sounds like a bug, we should either remove the Precondition test, or remove the begin() call if the EntityManager doesn't exist.

So, my suggestion would be either to fix the get() method, or change the documentation to tell about this potential issue, and suggest the injection of Provider&lt;EntityManager&gt; in every cases.

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=739_

asharayev:
I have exactly the same concern as @mikehearn has.
The problem is that in the get() method of JpaPersistService there is actually the check we want:
see lines 53-56:
```java
    EntityManager em = entityManager.get();
    Preconditions.checkState(null != em, "Requested EntityManager outside work unit. "
        + "Try calling UnitOfWork.begin() first, or use a PersistFilter if you "
        + "are inside a servlet environment.");
```
But, two lines up there is another code that makes the checkState() call ridiculous:
```java
    if (!isWorking()) {
      begin();
    }
```

I see that this is definitely a bug. Could it be fixed?

krdlab:
[...]

Issue #730, opened 2014-07-07
Automatically started UnitOfWork is never ended


gissuebot:
_From [spootsy.ootsy](https://code.google.com/u/103989710404024834643/) on September 26, 2012 22:48:16_

We've been using Guice in our projects for a while now, and it's been great. There's one feature/bug that is a reliable source of bugs for us, though.

When using the JpaPersistService, if you attempt to access an EntityManager outside of an active UnitOfWork, Guice will automatically start one for you. However, as Guice does not (and cannot) know when to end this UnitOfWork, it never does.

Result? The offending thread will be stuck with the same EntityManager throughout the life of the application. This is a bad state for the application to run in, and ours inevitably exhaust the available memory after a while and crash.

The real killer here is that it's not at all obvious when you've made this mistake. The only real tip-off is that you're getting inconsistent data from your database between different threads (due to the EMs first-level cache) or that the applications memory consumption keeps on growing.

I think it would be much better behaviour to simply throw an exception if an attempt is made to access an EntityManager instance outside of an active UnitOfWork. This lets the developer know that they've made a mistake and are attempting to access the database outside of a valid scope. Steps to reproduce: 1. 'mvn test' on the supplied maven project (assumes postgres running on localhost:5432 with DB/user/pass 'test')

**Binary attachments:** [guice_persist_same_em.zip](http://code.google.com/p/google-guice/issues/detail?id=730)

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=730_

woru:
+1 IMHO this a serious bug. You forget to add @Transactional and end up with OutOfMemoryError in production.
Also JpaPersistService is not public and cannot be overridden.

PR #997, opened 2016-02-28
Check a system property before EntityManager injection outside work unit happens

nuzayats:
There are many folks who complains about implicit EntityManager injection because it very likely to bring EntityManager leakage:

https://github.com/google/guice/issues/739
https://github.com/google/guice/issues/730

I think it's reasonable to check a system property before implicit injection happens. it has no compatibility breakage, and simple.

Inspired by workaround of this: https://issues.apache.org/jira/browse/COLLECTIONS-580


lbtc-xxx:
I signed it!

lbtc-xxx:
@sameb
Hi, I guess you might thought this PR wouldn't be a correct solution to fix this issue. If so it's fine but anyway I believe the issue which this PR addressed should be fixed.

Why don't you make injection process of `EntityManager` to throw Exception when it happens outside unit of work as the message in Exception states? Do you afraid of compatibility breakage?

Any feedback or thoughts greatly appreciated :)

PR #1014, opened 2016-05-16
Use LinkedList to store Elements in InjectorShell

biran0079:
Elements in the list are processed by various AbstractProcessor objects, and get removed when processing is all done.
Removing elements from an array list takes O(n) time, while it is O(1) for linked list. This can become bottleneck
for injector creation.

### DIFF
--- a/core/src/com/google/inject/internal/InjectorShell.java
+++ b/core/src/com/google/inject/internal/InjectorShell.java
@@ -68,7 +68,7 @@ final class InjectorShell {
   }
 
   static class Builder {
-    private final List<Element> elements = Lists.newArrayList();
+    private final List<Element> elements = Lists.newLinkedList();
     private final List<Module> modules = Lists.newArrayList();
 
     /** lazily constructed */


tavianator:
The removal happens [here](https://github.com/google/guice/blob/9f0fb561a5cfa269ee43b1503e3dd75d1493e9aa/core/src/com/google/inject/internal/AbstractProcessor.java#L58).  So this seems like a good change, a bunch of `Iterator::remove()` calls on an `ArrayList` seems bad.

But `LinkedList` has fair amount of overhead on its own.  Can we just not remove the elements at all?  The processors seem to only pay attention to the appropriate elements anyway, I don't think anything would get double-processed, but I haven't actually tested this theory.

@sameb?

tavianator:
If needed, a separate `boolean[] processed` to keep track of indices that shouldn't be re-visited would work, and would probably be faster than `LinkedList`.

sameb:
@lukesandberg -- may be a good perf opportunity here.

biran0079:
Thanks for looking at this!

The memory usage of `LinkedList` is probably similar to `ArrayList` + `boolean[] processed`.
`LinkedList` might have worse memory locality, while lazy deletion approach requires unnecessary iteration and boolean check, which is not ideal as well.
From simplicity point of view, `LinkedList` is definitely the easiest solution.

Overall I think `LinkedList` should be used here.

scottnice:
LinkedList also has poorer traversal performance than arraylist does. Switching to a linkedlist will mean faster deletion but the poor collection traversal will outweigh any performance benefits from deleting.

lukesandberg:
Do you have any numbers to suggest that this is actually faster?

sameb:
This was fixed in https://github.com/google/guice/commit/363f568b6de8c093d6669b45d29c804783c9f39e to use `removeIf`, which does not shift elements until all elements are processed.

PR #1015, opened 2016-05-18
Fix errornous value in issueManagement.system

jamo:
Just something small I noticed.


sameb:
It looks like this was fixed in https://github.com/google/guice/commit/3f5ce3f73cbf6fe1548a4718c0c01b9b888ab3f9.

PR #1023, opened 2016-06-23
Added xml and vis.js graphing extensions.

ksaric:
Simplified adding new graphing extensions, added xml and (simple) vis.js graphing.


ksaric:
6 month later - @sameb, did anyone took a look at this? I could fix the conflicting file, but to what purpose, if no one is going to check this out?

sameb:
Hello!  Much time has passed since this PR was raised. Thank you for the PR, however Guice's grapher extension is not an extension we're looking to expand.

PR #1024, opened 2016-06-24
Make Gerrit work on IBM's WLP

### LINKED ISSUE #1017: Bug in anonymous sub class of HttpServletRequestWrapper in doService()?
HI,

I thinks there is maybe a bug in the method getPathInfo() which calls getServletPath().
The problem is as we don't have any real Servlets here and only one real filter is installed in the servlet container, so getServletPath() will always return an empty string according to the documentation ( https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getServletPath-- ) because eveything wll be matched against /*

But guice-servlet users like expect the getServletPath to return something different as this call is made in computePath():
path = patternMatcher.extractPath(servletPath);

This call makes only sense when not an empty string is returned.

This broken behaviour makes it currently impossbile to run gerrit on any other servlet container than jetty 9.2 I guess.

thomasmey:
I thinks there is a bug in the method getPathInfo() which calls getServletPath().
The problem is as we don't have any real Servlets here and only one "real" filter
is installed in the servlet container, so getServletPath() will always return
an empty string according to the documentation ( https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getServletPath-- )
because eveything wll be matched against /*

But guice-servlet users like gerrit expect the getServletPath to return something
different as this call is made in computePath():

path = patternMatcher.extractPath(servletPath);

This call makes only sense when not an empty string is returned.

This broken behaviour makes it currently impossbile to run gerrit on any other
servlet container than jetty 9.2 I guess which seems to return something different
for above getServletPath() call. But I didn't check the concrete value that
jetty does return.

Fixes #1017

### DIFF
--- a/extensions/servlet/src/com/google/inject/servlet/ServletDefinition.java
+++ b/extensions/servlet/src/com/google/inject/servlet/ServletDefinition.java
@@ -261,6 +261,9 @@ class ServletDefinition implements ProviderWithExtensionVisitor<ServletDefinitio
       private String computePath() {
         if (!isPathComputed()) {
           String servletPath = super.getServletPath();
+          if(servletPath.isEmpty()) {
+              servletPath = super.getRequestURI().substring(getContextPath().length());
+          }
           path = patternMatcher.extractPath(servletPath);
           pathComputed = true;
 
--- a/extensions/servlet/src/com/google/inject/servlet/UriPatternType.java
+++ b/extensions/servlet/src/com/google/inject/servlet/UriPatternType.java
@@ -168,7 +168,7 @@ public enum UriPatternType {
       }
       return null;
     }
-    
+
     public UriPatternType getPatternType() {
       return UriPatternType.REGEX;
     }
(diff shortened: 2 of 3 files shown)


sameb:
In addition to signing the CLA, please add tests to the PR.  Thanks!

thomasmey:
I signed it!

thomasmey:
Hi, i changed author and committer, but the bots seemed to fail to pick up the change. Branch work on my Repo should be okay. Anything I can do to fix this?

thomasmey:
I also fixed the existing unit tests. Hope this is enough to get this into master

sameb:
could you maybe close this PR & re-open a new one w/ the correct author info in its initial pass to make the cla bot happy?  if that doesn't do it i can look to manually confirm.

hashar:
The new PR is #1025

PR #1025, opened 2016-06-29
Make Gerrit work on IBM's WLP

thomasmey:
I think there is a bug in the method getPathInfo() which calls
getServletPath().
The problem is as we don't have any real JSR-340-Servlets in gerrit;
instead only one "real" filter is installed in the Servlet container.
So getServletPath() will always return an empty string according to the
documentation (
https://docs.oracle.com/javaee/7/api/javax/servlet/http/HttpServletRequest.html#getServletPath--
)
because everything will be matched against /*

But guice-servlet users like gerrit expect the getServletPath() to
return something different as this call is made in computePath():
path = patternMatcher.extractPath(servletPath);

This call makes only sense when not an empty string is returned.

This broken behavior makes it currently impossible to run gerrit on any
other Servlet container than Jetty 9.2. For the URI
"http://localhost:9080/gerrit/projects/?n=26&type=ALL&d" Jetty does
return
"/projects" when getServletPath() is called. Which actually violates the
Servlet spec, I think.

For the reference: The Gerrit guice-servlet bindings are created in
gerrit-httpd/src/main/java/com/google/gerrit/httpd/UrlModule.java

### DIFF
--- a/extensions/servlet/src/com/google/inject/servlet/ServletDefinition.java
+++ b/extensions/servlet/src/com/google/inject/servlet/ServletDefinition.java
@@ -265,14 +265,16 @@ class ServletDefinition implements ProviderWithExtensionVisitor<ServletDefinitio
           private String computePath() {
             if (!isPathComputed()) {
               String servletPath = super.getServletPath();
+              if(servletPath.isEmpty()) {
+                servletPath = super.getRequestURI().substring(getContextPath().length());
+              }
               path = patternMatcher.extractPath(servletPath);
               pathComputed = true;
 
               if (null == path) {
-                path = servletPath;
+                  path = servletPath;
               }
             }
-
             return path;
           }
         };
--- a/extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
+++ b/extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
@@ -91,6 +91,12 @@ public class ServletDefinitionPathsTest extends TestCase {
 
     expect(request.getServletPath()).andReturn(requestPath);
 
+    // computePath()
+    if(requestPath.isEmpty()) {
+        expect(request.getRequestURI()).andReturn(requestPath);
+        expect(request.getContextPath()).andReturn("");
+    }
+
     replay(injector, binding, request);
 
     ServletDefinition servletDefinition =
@@ -194,6 +200,13 @@ public class ServletDefinitionPathsTest extends TestCase {
 
     expect(request.getServletPath()).andReturn(servletPath).anyTimes();
 
+    // computePath()
+    if(servletPath.isEmpty()) {
+        expect(request.getRequestURI()).andReturn(requestUri);
+        expect(request.getContextPath())
+        .andReturn(contextPath);
+    }
+
     expect(request.getContextPath()).andReturn(contextPath);
 
     expect(request.getAttribute(REQUEST_DISPATCHER_REQUEST)).andReturn(null);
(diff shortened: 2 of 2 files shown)


sameb:
Looks like the CLA still doesn't recognize you.  I also manually checked your github username and it doesn't see you, however it does recognize your email address.  Can you please update your CLA with your github username?

thomasmey:
I signed it!

extensions/servlet/src/com/google/inject/servlet/ServletDefinition.java
```diff
@@ -261,6 +261,9 @@ public String getPathTranslated() {
       private String computePath() {
         if (!isPathComputed()) {
           String servletPath = super.getServletPath();
+          if(servletPath.isEmpty()) {
```
sameb: style nits: please put a space between `if` and `(`

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
```diff
@@ -93,6 +93,12 @@ protected void service(HttpServletRequest servletRequest,
     expect(request.getServletPath())
         .andReturn(requestPath);
 
+    // computePath()
+    if(requestPath.isEmpty()) {
+        expect(request.getRequestURI()).andReturn(requestPath);
```
sameb: style nits: please indent 2 spaces for new lines (4 spaces for continuations of prior lines)

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
```diff
@@ -93,6 +93,12 @@ protected void service(HttpServletRequest servletRequest,
     expect(request.getServletPath())
         .andReturn(requestPath);
 
+    // computePath()
```
sameb: rm rather than commenting out.  but my main question here is: why is this no longer necessary?

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
```diff
@@ -192,6 +198,13 @@ protected void service(HttpServletRequest servletRequest,
         .andReturn(servletPath)
         .anyTimes();
 
+    // computePath()
+    if(servletPath.isEmpty()) {
+        expect(request.getRequestURI()).andReturn(requestUri);
+        expect(request.getContextPath())
+        .andReturn(contextPath);
```
sameb: (this is an example of where you'd use a 4 space indent since it's a continuation)

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
```diff
@@ -216,35 +229,39 @@ protected void service(HttpServletRequest servletRequest,
...
+    pathInfoWithRegexMatching("/path/thing/stuff.html", "/path", "/thing(/).*", "/stuff.html",
         "/thing");
-    pathInfoWithRegexMatching("/path/thing", "/path", "/thing/(.)*", null, "/thing");
+    pathInfoWithRegexMatching("/path/thing", "/path", "/thing(/).*", null, "/thing");
 
     // *.xx style mapping
-    pathInfoWithRegexMatching("/path/thing.thing", "/path", ".*\\.thing", null, "/thing.thing");
-    pathInfoWithRegexMatching("/path///h.thing", "/path", ".*\\.thing", null, "/h.thing");
-    pathInfoWithRegexMatching("/path///...//h.thing", "/path", ".*\\.thing", null,
+    //FIXME: what sense make the regex here anyway?
```
sameb: can you elaborate?  is this something wrong with the existing setup, or with the thing you're introducing?

extensions/servlet/src/com/google/inject/servlet/ServletDefinition.java
thomasmey: will fix

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
thomasmey: this line is completely new. the below code resembles the newly introduced logic in computePath() method.

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
thomasmey: I think I have the wrong formatter setting in eclipse :-)

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
```diff
@@ -216,35 +229,39 @@ protected void service(HttpServletRequest servletRequest,
 
   // Data-driven test.
   public final void testPathInfoWithRegexMatching() throws IOException, ServletException {
+
     // first a mapping of /*
-    pathInfoWithRegexMatching("/path/index.html", "/path", "/(.)*", "/index.html", "");
-    pathInfoWithRegexMatching("/path//hulaboo///index.html", "/path", "/(.)*",
+    pathInfoWithRegexMatching("/path/index.html", "/path", "(/).*", "/index.html", "");
```
sameb: IIUC, there's no change between these two regexs -- the former put a group around the "match any character" so the * was for the group.  the latter puts the group around the "/".  why did you change this?

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
```diff
@@ -216,35 +229,39 @@ protected void service(HttpServletRequest servletRequest,
...
-    pathInfoWithRegexMatching("/path/thing.thing", "/path", ".*\\.thing", null, "/thing.thing");
-    pathInfoWithRegexMatching("/path///h.thing", "/path", ".*\\.thing", null, "/h.thing");
-    pathInfoWithRegexMatching("/path///...//h.thing", "/path", ".*\\.thing", null,
+    //FIXME: what sense make the regex here anyway?
+    pathInfoWithRegexMatching("/path/thing.thing", "/path", "", null, "/thing.thing");
+    pathInfoWithRegexMatching("/path///h.thing", "/path", "", null, "/h.thing");
+    pathInfoWithRegexMatching("/path///...//h.thing", "/path", "", null,
         "/.../h.thing");
-    pathInfoWithRegexMatching("/path/my/h.thing", "/path", ".*\\.thing", null, "/my/h.thing");
+    pathInfoWithRegexMatching("/path/my/h.thing", "/path", "", null, "/my/h.thing");
```
sameb: why did this change?

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
sameb: ah sorry missed that it's new.  i'd use a comment with english if you want to explain that this is similar to something else, rather than commented out code that looks like it's stale.

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
thomasmey: yes, as far as I understand the existing tests are broken for the passed in regular expression. I think nearly none of those actually did match anything. So the implementation did fall back ( path == null -> use servletPath, see https://github.com/thomasmey/guice/blob/b2bcfaa3cc5ffaa74f0eed2a6e190ce30aae8460/extensions/servlet/src/com/google/inject/servlet/ServletDefinition.java#L271 )
With my fix, a correct servletPath is returned now with uncovers the broken tests.
At least that is what I hope :-)

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
thomasmey: I think it goes like this:

the regex pattern matcher returns the remaining path "behind - 1" the last group.
before my fix servletPath = "" - matched against /(.)* did return nothing because no match was found.

after my fix servlerPath is set to the reuqest URI without the context path. serlvetPath = "/index.html" -> matches against /(.)* and returns "l".

This is why I'm pretty sure the given regular expression in the unit tests are wrong, especially for all test cases with servletPath = "".

What do you think? I'm I right or do I have a thinko here?

extensions/servlet/test/com/google/inject/servlet/ServletDefinitionPathsTest.java
thomasmey: because a regular expression without group in the regular expression matcher makes no sense at all. The regular expression matcher doesn't work very intuitive. See the logic in RegexUriPatternMatcher.

thomasmey:
Hi Sam,

any update on this?

thomasmey:
I guess this is another dead open source project then?!

thomasmey:
Hi,

I rebased my branch on top of current master. See fae2cdb in above time line. sadly the rebase doesn't trigger a CI build.

So, would be cool if somebody could have a look at this PR and may pull it.

hashar:
The pull request is for issue #1017

sameb:
Hello. Much time has passed, and I'm not sure if this PR is still valid. I unfortunately don't have much faith in the unit tests for the servlet extension, and do not want to submit changes to the extension that seem to materially change the way the tests work. If someone would like to revive this PR, and wants to include integration tests that actually spin up a real servlet container and do E2E tests using Guice (rather than unit tests that mock out the behavior and quite possibly get it wrong), I would be more than happy to take another look at this. Until integration tests exist, I'm unfortunately not comfortable with these kinds of changes.

PR #1038, opened 2016-09-07
merge internal changes

sameb:


tavianator:
Saddened by the deprecation of `getDependencyChain()`, it's key to Sangria's `ContextSensitiveProvider` [implementation](https://github.com/tavianator/sangria/blob/master/sangria-contextual/src/main/java/com/tavianator/sangria/contextual/ContextSensitiveBinder.java#L339), which in turn enables first-class injection of non-jul loggers.  We also use something like it internally for lots of other things.

Is it possible to save some performance by filling in the dependency chain lazily?  It seems like that information already has to be available for constructing error messages anyway.

I know I can switch to the `ThreadLocal` thing, but it seems like intercepting literally every provision would be much worse from a performance standpoint.

sameb:
+ @lukesandberg 

Is the workaround in the docs (using a ThreadLocal) not enough to keep things working?

sameb:
Oh sorry I missed your last point.  Can you do some benchmarks to let us know if using the ThreadLocal to intercept everything makes it actually worse?  In our scenarios, folks that used getDependencyChain were already intercepting everything.

PR #1072, opened 2017-01-17
Fix error propagation in SingleMethodInjector

aadamowski:
Under certain circumstances (when errors.withSource() returns a new Errors
instance), the caught user exception can be lost and not propagated. This is
because the message from the cause gets added into the new Errors instance,
which gets subsequently discarded without being merged into the Errors instance
that is being processed by the SingleMethodInjector.inject() method.


aadamowski:
I signed it!

sameb:
Thank you for this PR. In the time since this PR was submitted, error handling has been significantly revamped. Specifically, as of https://github.com/google/guice/commit/f6cb899ce0787848eb86b32c82b741110774b6ed, there is no outer "errors" to add to anymore.

Issue #1075, opened 2017-01-31
Struts2Factory does not handle creation of Struts interceptors after creation of injector


ThrawnCA:
We have an application using the Guice Struts extension, which fails when we upgrade Struts to 2.5.6+

Debugging indicates that when using Struts 2.5.5 and below, `com.google.inject.struts.Struts2Factory` builds up its list of interceptors before calling `createInjector`, so they are populated properly. However, starting from Struts 2.5.6, the injector is created first, and the interceptors second, so `ProvidedInterceptor.delegate` is never populated, resulting in a null pointer failure in `ProvidedInterceptor.intercept`.

Perhaps the `buildInterceptor` method could detect whether or not the injector has been created (`strutsInjector != null`) and if so, inject the new interceptor immediately.

ThrawnCA:
This is now becoming critical, as S2-045 makes it urgent to update Struts to 2.5.10.1

tguless:
This is a duplicate of https://github.com/google/guice/issues/278

PR #1077, opened 2017-02-01
ensure that Struts interceptors are populated even if they are create…

ThrawnCA:
…d after the injector, github #1075


jatin039:
Signed it.

Issue #690, opened 2014-07-07
Do not fail on circular constructor/setter dependency


gissuebot:
_From [stepan.koltsov](https://code.google.com/u/113939464116033188989/) on March 07, 2012 14:42:18_
# 

public static class A {
&nbsp;&nbsp;&nbsp;&nbsp;private final B b;

&nbsp;&nbsp;&nbsp;&nbsp;`@`Inject
&nbsp;&nbsp;&nbsp;&nbsp;public A(B b) {
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;this.b = b;
&nbsp;&nbsp;&nbsp;&nbsp;}
}

public static class B {
&nbsp;&nbsp;&nbsp;&nbsp;private A a;

&nbsp;&nbsp;&nbsp;&nbsp;`@`Inject
&nbsp;&nbsp;&nbsp;&nbsp;public void setA(A a) {
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;this.a = a;
&nbsp;&nbsp;&nbsp;&nbsp;}
}

public static void main(String[] args) {
&nbsp;&nbsp;&nbsp;&nbsp;Guice.createInjector(new AbstractModule() {
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;`@`Override
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;protected void configure() { }
&nbsp;&nbsp;&nbsp;&nbsp;}).getInstance(A.class);
# }

guice reports error:
# 
# Tried proxying com.company.Temp$A to support a circular dependency, but it is not an interface

Proxy is not needed here: guice should just create "B" first, inject it into "A" constructor and then inject "A" into "B".

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=690_

marcosdeveloper:
I got this message for another link too: https://github.com/google/guice/issues/139.
First of all I think that the @Inject doesn't refer to concrete classes and yes, interfaces only.
I guess that the message have to improve to: 

> Tried proxying projectGuiceTest.A to support a circular dependency, but it is not an interface. You have to break the circular dependency, or make the circular dependency refers to interfaces.

I will made the changes to improve it.

Issue #139, opened 2014-07-07
Improve Circular Dependency Error Message


gissuebot:
_From [sberlin](https://code.google.com/u/sberlin/) on August 06, 2007 17:56:32_

Right now, a circular dependency produces the following error msg: 
com.google.inject.ConfigurationException: Tried proxying Circle to support 
a circular dependency, but it is not an interface.

It would be very useful if this message was improved to show who the 
players in the circular dependency are.&nbsp;&nbsp;That is, if class Circle depends 
on SquareHole and SquareHole depends on Circle, the message could say 
that's what it's failing.&nbsp;&nbsp;This would make fixing the problem much easier.

Here's a fuller stack trace, for reference: 

com.google.inject.ConfigurationException: Tried proxying Circle to support 
a circular dependency, but it is not an interface.
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.ConstructionContext.createProxy
(ConstructionContext.java:70)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.ConstructorInjector.construct
(ConstructorInjector.java:126)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at [... lots of internal guice calls ...]
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.InjectorImpl.getInstance
(InjectorImpl.java:728)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at [... my code ...]
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at sun.reflect.NativeMethodAccessorImpl.invoke
(NativeMethodAccessorImpl.java:39)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at sun.reflect.DelegatingMethodAccessorImpl.invoke
(DelegatingMethodAccessorImpl.java:25)
[...]

PR #1087, opened 2017-03-25
Do not fail on circular constructor/setter dependency #690

marcosdeveloper:
I got this message for another link too: #139.
First of all I think that the @Inject doesn't refer to concrete classes and yes, interfaces only.
I guess that the message have to improve to:

> Tried proxying projectGuiceTest.A to support a circular dependency, but it is not an interface. You have to break the circular dependency, or make the circular dependency refers to interfaces.

In this way I think that we clarify more the user who will make the best decision.

### DIFF
--- a/core/src/com/google/inject/internal/Errors.java
+++ b/core/src/com/google/inject/internal/Errors.java
@@ -603,7 +603,7 @@ public final class Errors implements Serializable {
 
   public Errors cannotProxyClass(Class<?> expectedType) {
     return addMessage(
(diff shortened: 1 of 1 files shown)


marcosdeveloper:
I signed it!

marcosdeveloper:
I signed it!

marcosdeveloper:
I signed it!

marcosdeveloper:
I signed it!

marcosdeveloper:
I signed it!

marcosdeveloper:
I signed it!

marcosdeveloper:
I signed it!

marcosdeveloper:
I signed it!

PR #1089, opened 2017-03-25
Do not fail on circular constructor/setter dependency #690

marcosdeveloper:
I got this message for another link too: #139.
First of all I think that the @Inject doesn't refer to concrete classes and yes, interfaces only.

I guess that the message have to improve to, as I did:

> Tried proxying projectGuiceTest.A to support a circular dependency, but it is not an interface. You have to break the circular dependency, or make the circular dependency refers to interfaces.

In this way I think that we clarify more the user who will make the best decision.

### DIFF
--- a/core/src/com/google/inject/internal/Errors.java
+++ b/core/src/com/google/inject/internal/Errors.java
@@ -603,7 +603,7 @@ public final class Errors implements Serializable {
 
   public Errors cannotProxyClass(Class<?> expectedType) {
     return addMessage(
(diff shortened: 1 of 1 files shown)


sameb:
Thank you for this PR. In the time since this PR was submitted, Guice has significantly changed it's error messaging, and these kinds of errors will now include a link to https://github.com/google/guice/wiki/CAN_NOT_PROXY_CLASS, which is more likely to help the user understand what to do.

PR #1095, opened 2017-04-07
Moe sync

dimo414:


dimo414:
FYI the unverified commit is from @ThrawnCA (who has signed) using a different email address.

PR #1108, opened 2017-06-02
Provide a way to create a copy of a Key with an annotation

kashike:
This allows you to easily specify an annotation/annotation type to an existing key.

### DIFF
--- a/core/src/com/google/inject/Key.java
+++ b/core/src/com/google/inject/Key.java
@@ -300,6 +300,24 @@ public class Key<T> {
     return new Key<T>(typeLiteral, annotationStrategy.withoutAttributes());
   }
 
+  /**
+   * Returns a new key of the same type with the specified annotation.
+   *
+   * @since 4.1
+   */
+  public Key<T> withAnnotation(Class<? extends Annotation> annotationType) {
+    return new Key<T>(typeLiteral, strategyFor(annotationType));
+  }
+
+  /**
+   * Returns a new key of the same type with the specified annotation.
+   *
+   * @since 4.1
+   */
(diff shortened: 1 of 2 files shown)


sameb:
Don't really think this is necessary. Trivial to do already:
   `Key.get(key.getTypeLiteral(), annotation)`

kashike:
It was to provide the opposite of `Key#ofType` which makes sense to have as it deals with the `AnnotationStrategy` itself.

There's not a huge difference with changing the annotation, one just involves less duplication:
`key.withAnnotation(annotation)`
`Key.get(key.getTypeLiteral(), annotation)`

If you don't think this is necessary, that's fine - feel free to close.

sameb:
ofType is slightly more necessary, as construction a new variant with a different type requires some branching (to deal with annotation instances vs annotation types).

dimo414:
Personally, the discoverability here is at least a bit of a selling point. More than once I've looked for an `ofType()` sibling for annotations in `Key` and realized I need to instead use the factory method.

sameb:
if you think it's worth it extra methods, feel free to merge in.

kashike:
Any chance of this being pulled soon, @dimo414?

kashike:
Any chance of getting this in soon? :smiley_cat:

kashike:
Is there a chance of this getting in soon?

dimo414:
I'll take a swing at it over the long weekend :)

kashike:
Thank you! :smile:

dimo414:
Thanks for the contribution, and your patience :)

PR #1109, opened 2017-06-02
Fix type mismatch in ProvisionListenerCallbackStore.remove

robertbrignull:
The type of the value returned from `cache.asMap()` is `Map<KeyBinding, ProvisionListenerStackCallback<?>`, however we were trying to remove a `Binding`. This would never work as the hashcode/equals for a `Binding` are different from a `KeyBinding`. I have no experience with this code but it seems all that is necessary is wrapping it as `new KeyBinding(type.getKey(), type)` so this is a very simple fix.

I added a test for this though maybe you consider this too trivial to test for in which case I can easily remove it from the PR, but the test does indeed fail before the change and pass afterwards so it seems useful.

This bug was found using [lgtm.com](https://lgtm.com/projects/g/google/guice/) where Guice is one of the many thousands of open source projects analysed. It turns out Guice has no other alerts which is very impressive, but there are many other projects that do.

### DIFF
--- a/core/src/com/google/inject/internal/ProvisionListenerCallbackStore.java
+++ b/core/src/com/google/inject/internal/ProvisionListenerCallbackStore.java
@@ -86,7 +86,7 @@ final class ProvisionListenerCallbackStore {
    * <p>Returns true if the type was stored in the cache, false otherwise.
    */
   boolean remove(Binding<?> type) {
-    return cache.asMap().remove(type) != null;
+    return cache.asMap().remove(new KeyBinding(type.getKey(), type)) != null;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


dimo414:
Nice find; I would have expected [Error Prone](https://github.com/google/error-prone) to catch this, but probably no one's touched this file in a while in order to notice.

For posterity, here's the [alert in context](https://lgtm.com/projects/g/google/guice/snapshot/dist-12160006-1495130160887/files/core/src/com/google/inject/internal/ProvisionListenerCallbackStore.java#L89), and the [error description](https://lgtm.com/rules/5990065/).

Issue #850, opened 2014-09-15
OutOfMemoryError  in com.google.inject.internal.Errors


michalradziwon:
Hello,

I am getting the OutOfMemoryError caused by a huge number of errors stored in com.google.inject.internal.Errors.errors array list (see the stacktrace at the end of the description).
All I want to do is to be able to define a provider of the List class that is being used in the Errors class. After introducing this feature, I will be able to pass the fixed-sized list to the Errors instance (for example max 1000 elements) and just throw an exception (with a list of first 1000 errors) when the size is exceded.
After a quick static code analysis I found that I can't provide my own Errors class implementation (it is final).

Here is a proposition of changes:
Modify com.google.inject.internal.Errors class:
1. add new Supplier< List< Message > > field - a reference to message list's factory (with bound default implementation, that returns new instance of Lists.newArrayList() on each call)
2. add new Supplier< List < Message > > field - a reference to message list's factory builder method that will allow to assign the message list's factory to the field

Modify com.google.inject.internal.InternalInjectorCreator class:
1. Add new constructor with one extra argument: Supplier< List < Message > > - message list's factory
2. Constructor should call InternalInjectorCreator.errors.withListProvider() method, passing the message list's factory

What do you think about that ?

com.google.inject.internal.util.$ComputationException: java.lang.OutOfMemoryError: Java heap space.
com.google.inject.internal.util.$ComputationException: java.lang.OutOfMemoryError: Java heap space
    at com.google.inject.internal.util.$MapMaker$StrategyImpl.compute(MapMaker.java:553)
    at com.google.inject.internal.util.$MapMaker$StrategyImpl.compute(MapMaker.java:419)
    at com.google.inject.internal.util.$CustomConcurrentHashMap$ComputingImpl.get(CustomConcurrentHashMap.java:2041)
    at com.google.inject.internal.FailableCache.get(FailableCache.java:50)
    at com.google.inject.internal.ConstructorInjectorStore.get(ConstructorInjectorStore.java:49)
    at com.google.inject.internal.ConstructorBindingImpl.initialize(ConstructorBindingImpl.java:125)
    at com.google.inject.internal.InjectorImpl.initializeJitBinding(InjectorImpl.java:521)
    at com.google.inject.internal.InjectorImpl.createJustInTimeBinding(InjectorImpl.java:847)
    at com.google.inject.internal.InjectorImpl.createJustInTimeBindingRecursive(InjectorImpl.java:772)
[...]

PR #1122, opened 2017-08-01
Added posibility to limit errors logger to prevent OOM

IgeNiaI:
In our project we encountered an issue where we got a lot of errors during guice injection. There were so much errors that the injector run into OutOfMemoryError when formatting error messages in Errors.format(String, Collection<>) method.
The behavior is pretty much the same as here: https://github.com/google/guice/issues/850

I have added the possibility to set limit for errors being logged.
In the code it can be used by just calling the Errors.setMaxErrorsLogged(int) method.


IgeNiaI:
I signed it!

PR #1127, opened 2017-09-19
Adding implementation for a module that will only be installed once

mveitas:
This is helpful for cases where you want to ensure that a module is only configured once. I wasn't quite sure of the name, open to suggestions

### DIFF
--- a/core/src/com/google/inject/SingletonModule.java
+++ b/core/src/com/google/inject/SingletonModule.java
@@ -1,0 +1,23 @@
+package com.google.inject;
+
+public abstract class SingletonModule extends AbstractModule {
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+
+        if (!(obj instanceof AbstractModule)) {
+            return false;
+        }
+
+        return getClass().equals(obj.getClass());
+    }
+
+    @Override
+    public int hashCode() {
+        return getClass().hashCode();
+    }
+
+}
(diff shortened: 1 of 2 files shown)


mveitas:
I signed it!

dimo414:
@mveitas thanks for the contribution! We actually have an internal class that does the same thing as your proposed module, though we've hesitated to open-source it since it takes advantage of (presently) undocumented behavior.

I would suggest filing a bug asking to add such a class to Guice, then we can evaluate whether that makes sense. The change itself is small, but choosing to make this implementation detail part of the contract is a bigger question.

kashike:
It's worth noting that a version of this class exists in quite a few projects that use Guice.
A couple examples:
- Governator: [here](https://github.com/Netflix/governator/blob/master/governator-core/src/main/java/com/netflix/governator/SingletonModule.java) and [here](https://github.com/Netflix/governator/blob/master/governator-legacy/src/main/java/com/netflix/governator/guice/SingletonModule.java)
- [guice-dedupe](https://github.com/fzakaria/guide-dedupe/blob/master/src/main/java/com/github/fzakaria/guice/dedupe/SingletonModule.java) (http://blog.fzakaria.com/2016/12/11/large-scale-guice-projects-module-deduplication/)

twicksell:
Speaker as a Governator owner, @dimo414 is correct in that this implementation does raise some larger questions. The behavior of this in combination with Modules.override for example is poorly defined. I'd definitely +1 for a solution to this issue, but this implementation by itself feels incomplete.

Issue #1128, opened 2017-09-21
Feature request: just-in-time binding resolution through SPI


pshirshov:
Let's imaging that I wish to inject some values which cannot be constructed ahead of time. For example, config values:

```
class Something @Inject() (@ConfigValue("x.y.z") value: MyValue) {...}
```

Right now I have no effective way to do that as far as I can understand.

Though in some situation such functionality would be priceless.

pshirshov:
Eh, guys. I've already wrote [my very own DI framework](/7mind/izumi) which outperforms Guice in almost any aspect but you didn't even consider accepting my P/R yet. :/

PR #1130, opened 2017-09-21
#1128 just-in-time instantiation through SPI

pshirshov:
Look at #1128  for motivation

### DIFF
--- a/core/src/com/google/inject/internal/InjectorImpl.java
+++ b/core/src/com/google/inject/internal/InjectorImpl.java
@@ -23,6 +23,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ListMultimap;
+import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import com.google.inject.Binder;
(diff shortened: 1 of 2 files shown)


pshirshov:
Eh, guys. I've already wrote my very own DI framework which outperforms Guice in almost any aspect but you didn't even consider accepting this branch yet. :/

GedMarc:
Lol @pshirshov Same here, already completed a full JPMS xD
Migrated the Microsoft SQL Server driver to JPMS, shit you not 40 minutes approved and merged.

It's like opposite world, since when does google not give a shit xD

GedMarc:
@pshirshov if possible, could you please try create a pull request on https://github.com/GedMarc/guice/tree/master ?
This  is fully up to date and JPMS 100% 

It's maven location is https://search.maven.org/artifact/com.jwebmp.inject/guice/0.67.0.4/jar

:)

sameb:
Thank you for the contribution, and sorry that you had to wait this unacceptably long time for a reply. The PR as proposed is not something we'd want to add to the Guice API, as it makes provisioning somewhat non-deterministic (based on the order it encounters a "custom provider", which isn't defined).

Even with a better API, I don't think we'd want to support such custom JIT provisioning, though I could possibly be convinced.

GedMarc:
Lol @pshirshov 
How to know you're a second grade developer, you rip off the DI running the entire of Google thinking a CDI or some other one is better...

Try not to rip off people's tools, you may actually be in the wrong.....
A thank you for the notice and hard work might have been more appreciative

pshirshov:
😉

sameb:
@all - please remember to be kind to each other. It's OK to have opinions on projects and/or disagree with them, but it's not OK to insult or attack individuals.

pshirshov:
@sameb : well, if you are still replying here, could you answer how do you imagine me convincing someone? Should wait six more years to convince then six more years to get the "proper" patch accepted? :smile:

pshirshov:
Eh, I this conversation happens back in 2017 I would show and explain, now I have zero motivation for this. I'm not sure why you call Guice "flexible", it won't allow me to resolve the problem in the original ticket nor, for example, dump a graphviz file effectlessly.

Still, I would like to add a couple of crticial comments. 

1. It doesn't matter how widely a library is adopted if it's outdated and badly designed. In my opinion, Guice is promotes extremely outdated practices, extremely badly designed (too ad-hoc; no planning; too many things happen just in time,;no resource management; too many things happen in runtime; no proper support for resource/lifecycle management, memoization, inheritance; invasive; nondetermenistic; unpredictable outside of happy paths; no sane conflict resolution; laziness instead of graph roots) and poorly implemented (unmaintainable overcoupled code full of singletons). A proper theory (project networks) allows to implement orders of magnitude more sound and determenistic DI.
2. I'm wondering how the world where Google won't use a subpar and extremely imperfect tech as one of the cornerstones would look like.

PR #1131, opened 2017-09-29
Applying FUNC transformation using Rascal

jmarcelonunes:
Using Rascal Software retrieved from https://github.com/refactoring-towards-language-evolution/rascal-Java8. FUNC transformation was applied and the result was the following log:

- Number of files:  556
- Processed Files: 21
- Exported Files:   10
- Total of files changed: 19
- Total of transformations: 35
- Errors: 45


ljacqu:
This is not very good; it breaks the formatting and replaces some for-each loops with a `Collection#forEach` lambda...?

sameb:
I'm not really sure what the purpose of this CL is, the only functional difference I see is changing `for(x : y) loops to `y.forEach(x -> {..})` calls. There's no reason to make that change as far as I can see.

Also: the CLA is not signed, and formatting is broken.

PR #1137, opened 2017-11-16
accomodate manually initiated rollbacks

### LINKED ISSUE #1136: JpaLocalTxnInterceptor incorrectly assumes that @Transactional methods never handle exceptions
When reading through `JpaLocalTxnInterceptor#invoke` we can see the following comment:

```java
   //everything was normal so commit the txn (do not move into try block above as it
    //  interferes with the advised method's throwing semantics)
    try {
      txn.commit();
    } finally {
      //close the em if necessary
      if (null != didWeStartWork.get() ) {
        didWeStartWork.remove();
        unitOfWork.end();
      }
    }
```

This is a patently false assumption when considering persistence methods that follow another schema. Example:

```
@Transactional
public NiceResponse<EntityType> create(EntityType instance) {
  try{
    try {
      entityManager.persist(instance);
      entityManager.flush();
      return NiceResponse.success(instance);
    } catch (PersistenceException ex) {
       // unwrap to find actual cause and use that to find out what happened
      entityManager.getTransaction().rollback(); // this is theoretically optional
      thow ex.getCause();
    }
  } catch (ConstraintViolationException ex) {
    return NiceResponse.failure("Some useful error message", ex);
  }
  // possibly further catch blocks
}
```

Commiting the transaction after this method **will** throw an Exception all error cases, simply because the transaction has been rolled back already, but there's no Exception thrown to early-exit `invoke`.

Vogel612:
When methods marked @Transactional initiated a rollback without
throwing an exception, the Interceptor threw an exception with
either "Transaction not active" or "Transaction marked for rollback only"..

Fixes #1136

### DIFF
--- a/extensions/persist/src/com/google/inject/persist/jpa/JpaLocalTxnInterceptor.java
+++ b/extensions/persist/src/com/google/inject/persist/jpa/JpaLocalTxnInterceptor.java
@@ -82,7 +82,13 @@ class JpaLocalTxnInterceptor implements MethodInterceptor {
     //everything was normal so commit the txn (do not move into try block above as it
     //  interferes with the advised method's throwing semantics)
     try {
-      txn.commit();
+      if (txn.isActive()) {
+        if (txn.getRollbackOnly()) {
+          txn.rollback();
+        } else {
+          txn.commit();
+        }
+      }
     } finally {
       //close the em if necessary
       if (null != didWeStartWork.get()) {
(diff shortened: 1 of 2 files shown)


Vogel612:
Build on jdk7 failed due to a configuration error, while this PR does not touch any CI-related configuration. Could someone please restart the build and check whether the problem is persistent?

Vogel612:
Poking this PR since there recently has been activity on the repository beyond just syncing from internal development. Could I please get some feedback on the state of this PR (and possibly the validity of the issue it's trying to solve?)

Chris-Lercher:
We've been eagerly waiting for this to get fixed. As mentioned by the OP, it seems, that the posted commit would work fine, and that the build error was only of temporary nature, completely unrelated to the commit.

This bug is quite serious, as it makes guice-persist virtually unusable if the DB connection is not extremely stable (i.e. in many real world situations).

PR #1141, opened 2017-12-08
Moe sync

ronshapiro:


ronshapiro:
The tests in maven are still failing, and I believe someone on the guice team knows why.

Apparently oraclejdk7 is no longer supported (https://github.com/google/jimfs/commit/8c43f933a993f1c420eab4d6e04affd4a6a6e0f6) so we should ignore those as well.

PR #1143, opened 2018-01-06
Compiles with oracle jdk8 and jdk9.

natros:
Compiles with maven 3.5.2 using oracle jdk8 and jdk9.


natros:
I signed it!

core/src/com/google/inject/internal/Errors.java
```diff
@@ -736,55 +702,16 @@ public int compare(Message a, Message b) {
...
-
-    if (errorMessages.size() == 1) {
-      fmt.format("1 error");
-    } else {
-      fmt.format("%s errors", errorMessages.size());
-    }
-
-    return fmt.toString();
-  }
+  // TODO(lukes): move @Provides logic into @Provides methods to help simplify this monster
```
GaspardPO: What do we do with theses TODO ? do we have an Open Issue for them?

core/src/com/google/inject/internal/Errors.java
```diff
@@ -716,11 +684,9 @@ public Errors addMessage(Message message) {
     return this;
   }
 
+  // TODO(lukes): inline into callers
```
GaspardPO: What do we do with theses TODO ? do we have an Open Issue for them?

Issue #1177, opened 2018-04-12
Expose and/or matchers


kashike:
The [and](https://github.com/google/guice/blob/master/core/src/com/google/inject/matcher/AbstractMatcher.java#L38) and [or](https://github.com/google/guice/blob/master/core/src/com/google/inject/matcher/AbstractMatcher.java#L71) matcher implementations are restricted to use by `AbstractMatcher`, unlike the [not](https://github.com/google/guice/blob/master/core/src/com/google/inject/matcher/Matchers.java#L63) (and other matchers in this class) matcher - would it be possible to expose these as `Matchers.and(a, b)` and `Matchers.or(a, b)`? This would allow use of these matchers without extending `AbstractMatcher` for the default implementation, or creating your own matchers.

cterse:
Its been quite long since this issue was opened. Can I look into this?

kashike:
@cterse see #1200

sameb:
Yeah, that sounds reasonable.

PR #1200, opened 2018-07-26
Expose and/or matchers

kashike:
#1177


kashike:
Any chance of getting this in soon? :smiley_cat:

sameb:
Nope, use composition and the existing and(..) or or(..) methods instead.

kashike:
....that doesn't solve the issue this was trying to fix - see #1177...

...though I did just realise that I missed including my additions to `Matchers`.

Issue #1198, opened 2018-07-13
Stop shading ASM?


ronshapiro:
Per https://github.com/google/guice/pull/1169#issuecomment-395095124, maybe it's time to stop shading ASM?

That will allow people to bring their own copy of ASM if they so choose, and will unblock them when new versions of the JDK come out and Guice hasn't updated yet. We'd obviously provide a default in our maven pom for those that aren't ahead of the game. This could be helpful as the JDK continues its 6-month release cadence.

sameb:
sgtm, i believe the reason it's shaded is because historically asm was very binary incompatible with prior releases... but i think they changed that as of recent releases.

mkurz:
[That comment](https://github.com/google/guice/pull/1169#issuecomment-395097383) from @mcculls might be relevant here as well:
> Note that the build already produces a version of Guice that doesn't bundle asm or cglib:
>
> http://repo1.maven.org/maven2/com/google/inject/guice/4.2.0/guice-4.2.0-classes.jar
>
> This is a jar of the Guice classes before any JarJar'ing.
>
> Also if you don't need AOP then there's the "no_aop" jar which doesn't use asm/cglib at all:
>
> http://repo1.maven.org/maven2/com/google/inject/guice/4.2.0/guice-4.2.0-no_aop.jar

cushon:
> unblock them when new versions of the JDK come out and Guice hasn't updated yet

I don't think thing it's sufficient to not shade ASM, since the API level will also need to be updated to support new class file versions:

https://github.com/google/guice/blob/35caaec1b4af0f20ab8b3fae8a1fc98afc0fd13e/core/src/com/google/inject/internal/util/LineNumbers.java#L144

davido:
> I don't think thing it's sufficient to not shade ASM, since the API level will also need to be updated to support new class file versions:[...]

And for experimental support of JDK11 in Guice it could only use experimental `Opcodes.ASM7_EXPERIMENTAL`: [1].

https://gitlab.ow2.org/asm/asm/blob/master/asm/src/main/java/org/objectweb/asm/Opcodes.java#L55

davido:
> And for experimental support of JDK11 in Guice it could only use experimental Opcodes.ASM7_EXPERIMENTAL: [1].

Done in this PR #1203.

mkurz:
BTW: asm 6.2.1 is available

sameb:
[...]

PR #1197, opened 2018-07-13
Moe Sync

ronshapiro:
This code has been reviewed and submitted internally. Feel free to discuss on the PR and we can submit follow-up changes as necessary.

Commits:
=====
<p> Update guice opensource asm dep to 6.2.

af1dc88a809bf21f6f7971c0216ec0ede7de0f7f

Fixes https://github.com/google/guice/pull/1169

PR #1169, opened 2018-03-28
Bump asm version to 6.2 for JDK 10/11 support

svrakitin:

PR #1203, opened 2018-08-26
Jdk11 compatibility

davido:
Switch to experimental asm API that supports JDK11.

### DIFF
--- a/core/src/com/google/inject/internal/util/LineNumbers.java
+++ b/core/src/com/google/inject/internal/util/LineNumbers.java
@@ -141,7 +141,7 @@ final class LineNumbers {
     private String name;
 
     LineNumberReader() {
-      super(Opcodes.ASM6);
+      super(Opcodes.ASM7_EXPERIMENTAL);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


cushon:
What are the compatibility implications of this? The docs [1] for `ASM7_EXPERIMENTAL` say:

> Experimental, use at your own risk. This field will be renamed when it becomes stable, this will break existing code using it.

So there's going to be a source compatibility issue, but I'm not sure if it's also a binary compatibility issue: `ASM7_EXPERIMENTAL` is a `static final` field that will be inlined, but will future ASM releases continue to accept that option, or will the eventual non-experimental `ASM7` have a different value?

[1] https://gitlab.ow2.org/asm/asm/blob/3f7d6ba206ec6280a584bc135cd86c2b2cf5c279/asm/src/main/java/org/objectweb/asm/Opcodes.java#L50

mkurz:
@davido I think just updating the `pom.xml` isn't enough: You also have to upgrade the `jar` file, see how that was done earlier: #1197 (and also #1169)

davido:
@mkurz, Ah thanks for pointing this out. Apparently Guice is supporting both build tools Apache Ant and Apache Maven? I tested this PR with Maven and it worked for Gerrit Code Review purpose (Build and run with JDK11).

frankgrimes97:
OpenJDK 11 has now been released... any ETA on when we might expect this to be addressed?

mkurz:
@frankgrimes97 See these comments here: https://github.com/google/guice/issues/1198#issuecomment-423221773 - Seems like we get an updated Guice very soon.

PR #1209, opened 2018-10-03
Remove very out of date parent POM

elharo:
Remove very out of date parent POM that does not supply any current and useful data.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,12 +17,6 @@ See the Apache License Version 2.0 for the specific language governing permissio
 
   <modelVersion>4.0.0</modelVersion>
 
-  <parent>
-    <groupId>com.google</groupId>
-    <artifactId>google</artifactId>
-    <version>5</version>
-  </parent>
-
   <packaging>pom</packaging>
 
   <groupId>com.google.inject</groupId>


sameb:
Can you elaborate on why you're making this change, and how you validated it doesn't supply any current or useful data?  AFAICT, "5" is the most recent version of the Google parent POM (see http://repo1.maven.org/maven2/com/google/google/ -- no versions after 5).

Also, per http://repo1.maven.org/maven2/com/google/google/5/google-5.pom, it seems to fill in a few fields that the guice parent POM doesn't:
  ```
<developers>
  <developer>
    <id>google</id>
  </developer>
</developers>
```
&

```
<profiles>
  <profile>
    <id>release-sign-artifacts</id>
    <activation>
      <property>
        <name>performRelease</name>
        <value>true</value>
      </property>
    </activation>
    <build>
      <plugins>
        <plugin>
          <groupId>org.apache.maven.plugins</groupId>
          <artifactId>maven-gpg-plugin</artifactId>
          <executions>
            <execution>
              <id>sign-artifacts</id>
              <phase>verify</phase>
              <goals>
                <goal>sign</goal>
              </goals>
            </execution>
          </executions>
        </plugin>
      </plugins>
    </build>
  </profile>
</profiles>
```

elharo:
To some extent it depends on which side of Chesterton's fence you like to live on. I prefer the, "I don't know that fence is there (and the tests pass) so let's take it down." This came up because another project I'm more familiar with this cited guice as an example of why they needed this parent. I.e. people are cargo culting this parent. I'm sure someone had a plan for it back in 2010 but no one has touched it since, and  there's probably not even an owner for it now. 

The developer element is very unlikely to matter to anyone and, now that my attention is called to it, violates our current branding and trademark guidelines. In 2018 maven-gpg-plugin shouldn't be needed either. We should be using kokoro to sign the artifacts with Google's keys.

sameb:
I'm grateful you assume we're keeping on top of the latest trends... but we aren't. I have no idea what Kokoro is, and I'm pretty sure we don't use it to sign the release.  I sign the releases with the gpg plugin and my sameb@google.com account, which I'm pretty sure is the same way Guava and folks do it too (except with a different signer).

elharo:
OK, signing is a separate issue we should also address. tldr; we should not be signing with your corp email address. More details internally if you like. I'd be very surprised if Guava is still doing that. 

@cpovirk Do you happen to know how Guava is signing its jar these days?

elharo:
$  jarsigner -verify guava-26.0-jre.jar

jar is unsigned.

OK, let me push that onto the stack of things we need to fix too.

sameb:
Absent any further info, this isn't going to change.

Issue #972, opened 2015-12-29
JpaPersistService.begin() throws exception if a Unit of Work is already opened


jccarrillo:
The implementation of UnitOfWork.begin() within JpaPersistService does not comply with the definition specified in the javadoc.

`
"If there is already one open, the invocation will do nothing. In this way, you can define arbitrary units-of-work that nest within one another safely."
`

The above is not true, the implementation throws an exception if "one" is already opened.

skuzzle:
Duplicate of #947. Thanks for the fix. I hope it will be accepted.

Issue #598, opened 2014-07-07
Persist Extension: PersistService.start() cannot be called multiple times


gissuebot:
_From [michal.galet](https://code.google.com/u/111108595136478275222/) on February 02, 2011 04:06:48_

The PersistService.start() method claims that it is possible to call this method multiple times. The implementation however throws "java.lang.IllegalStateException: Persistence service was already initialized." on second call.

This is a problem since it is not possible to determine if the service was already started. Moreover the PersistFilter starts the PersistService in initialization. Therefore it is not possible to use the persistence in any ServletContextListener!

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=598_

lextiz:
I want to connect to the database straight after I have configured the JpaPersistModule instance, but this is not possible until servlet container initializes the PersistFilter, this means there is no option to load some data from DB and use it in @Provider methods. I think that there should be at least an option to configure the filter to initialize itself before servlet initialization stage of servlet container. Or option to start the JpaPersistModule explicitly, so that PersistFilter would not fail on init.

jhalterman:
@dhanji It is simple to create a new injector, but when unit testing and the injector takes a few seconds to create, the cost adds up quick.

dhanji:
@jhalterman I agree, but what is the purpose of restarting the PersistService? If it's just clearing out the session data, you can do that by shutting down the sessions and clearing L2 caches. Or clearing out the DB tables underneath.

jhalterman:
I'm not working on this now, but [some tests](https://github.com/apache/ambari/blob/trunk/ambari-server/src/test/java/org/apache/ambari/server/controller/AmbariManagementControllerTest.java) I was looking at generally spent a lot of time setting up a PersistService/EntityManager - a few seconds for each test case. I'm not sure how much of that could actually be saved by reusing an EntityManager/PersistService without having state from one test case interfere with another, but that's basically what I was aiming to do. And that's where I was aiming to have PersistService be generally more reusable.

fsparv:
FYI this seems to cause issues with guice-persist http://stackoverflow.com/questions/17402081/how-to-start-jpa-in-a-guice-quartz-web-application

fsparv:
also http://stackoverflow.com/questions/20337037/guice-creationexception-with-persistfilter... and I'm experiencing something similar right now...

PR #1214, opened 2018-10-12
Update JpaPersistService to ignore subsequent calls to start() and stop()

### LINKED ISSUE #947: Documentation of PersistService does not match its implementation
The JavaDoc on its `start()` and `stop()` method states that nothing should happen if service is already started or stopped respectively. However, trying to call those methods twice results in an IllegalStateException being thrown by the implementation due to these checks:

```java
public synchronized void start() {
    Preconditions.checkState(null == emFactory, "Persistence service was already initialized.");
    ...
}
```

```java
public synchronized void stop() {
    Preconditions.checkState(emFactory.isOpen(), "Persistence service was already shut down.");
    ....
```

chrisparton1991:
This resolves #947 and supersedes #972 and #598. It possibly also supersedes #1132, although I'm not sure what that user's use case is.

The current behaviour is to throw an `IllegalStateException` when `JpaPersistService.start()` or `JpaPersistService.stop()` are called multiple times. However, this clashes with the javadoc in `PersistService` and `UnitOfWork`, both of which state that subsequent calls to these methods are no-ops.

`JpaPersistService.end()` already handles multiple calls, and has the following comment:
```// Let's not penalize users for calling end() multiple times.```

This PR brings `JpaPersistService` and related tests into line with the documentation and resolves a pain point that multiple people (including myself) have experienced, e.g. https://stackoverflow.com/questions/17402081/how-to-start-jpa-in-a-guice-quartz-web-application.

Thank you for taking the time to look at my PR.

### DIFF
--- a/extensions/persist/src/com/google/inject/persist/jpa/JpaPersistService.java
+++ b/extensions/persist/src/com/google/inject/persist/jpa/JpaPersistService.java
@@ -103,7 +103,9 @@ class JpaPersistService implements Provider<EntityManager>, UnitOfWork, PersistS
 
   @Override
   public synchronized void start() {
-    Preconditions.checkState(null == emFactory, "Persistence service was already initialized.");
+    if (null != emFactory) {
+      return;
+    }
 
     if (null != persistenceProperties) {
       this.emFactory =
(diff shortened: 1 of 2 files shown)


jlannoy:
Thanks Chris ; I came here to fix the same problem... And saw you already did the job. I hope we'll get a release soon :)

chrisparton1991:
@jlannoy You're more than welcome, I hope we do too :) The PR got approved along with a few others recently, so we'll see what happens!

PR #1217, opened 2018-10-22
Moe Sync

ronshapiro:

Commits:
=====
<p> Suppress warnings in classes that implement equals() without also implementing
hashCode().

The contract for Object.hashCode states that if two objects are equal, then
calling the hashCode() method on each of the two objects must produce the same
result. Implementing equals() but not hashCode() causes broken behaviour when
trying to store the object in a collection.

848dce9584e6cf1560241368f3fcf5562aaa7964


ronshapiro:
Travis seemed to have missed this PR during Github's partial outage yesterday. Closing and reopening with the hopes that it will run the tests.

PR #1219, opened 2018-11-05
Add oraclejdk-ea to Travis

gliptak:


sameb:
can you clarify why you're adding this and what this does?

PR #1221, opened 2018-11-08
Moe Sync

ronshapiro:

Commits:
=====
<p> Add the @Immutable/@ThreadSafe annotations to classes used in Apps Framework’s Producers or RequestScope. Both annotations enforce at compile-time that the annotated class meets the annotation’s requirements.

This migration is being done to clearly document the requirements of
the annotated classes and so Apps Framework can simplify its code for
checking the requirements at server startup.

3dd67e6c08a0c904d227efa3134c9ac78a1949c0

### DIFF
--- a/core/src/com/google/inject/util/Modules.java
+++ b/core/src/com/google/inject/util/Modules.java
@@ -22,6 +22,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.errorprone.annotations.Immutable;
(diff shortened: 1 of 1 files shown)


ronshapiro:
@sameb all of the `ant` builds are failing. Perhaps the `@Immutable` annotation is in another jar that's not yet included there?

sameb:
blargh, sorry-- i converted our internal tests to use mvn for checking opensource compatibility since that's more exhaustive, but i guess we still need _some_ ant coverage.  i'll fix this.  (really we should just kill ant entirely, but i need to figure out how to do the javadocs & jdiff w/ maven to do that.)

PR #1242, opened 2019-04-25
Fix a typo: s/eligile/eligible

schwabj:


dimo414:
Looks like this was fixed in 6f1c6ccdd. Thanks for sending a PR.

PR #1244, opened 2019-05-05
Moe Sync

ronshapiro:

Commits:
=====
<p> Fix @ProvidesIntoSet reference

e2d7db78bc1b9a4f4b2925d902320c54d395ecb7

-------

<p> Create an immutable snapshot of the modules provided to DaggerAdapter.from

57bfb02d9ccfce0ca6be41ed6d34c572a0f0cf49

-------

<p> Validate that no unsupported annotations exist on modules from DaggerAdapter usages

b8e70ba02046a3dd2b2ad55b5e7fe6479cb83ca9

-------

<p> change to openjdk8/9 since travis likes those better now.

d42adbe71cc710625f6555ae5d5deba79de942ad

-------

<p> Add support for passing class literals to DaggerAdapter (and therefore ProviderMethodsModule).

This allows static @Provides methods to work correctly for interface modules. static @Provides + interface is not a huge win, but similar support will be necessary when we add support for abstract binding methods.

This temporarily avoids using FastClass for static methods when a class literal is passed due to a bug in cglib. https://github.com/cglib/cglib/pull/153 is out to fix that, at which point we can enable FastClass support as usual.

0b77bc04a93491a245c367eda6143ef4caa9a315

-------

<p> No longer use Arrays.asList() now that we're using an ImmutableList

e0470189b524cd992b174edd7f39f9f91c2892e5

-------

<p> Compile against the newest version of Dagger for dagger-adapter, and make it optional so that users must bring their own version of dagger.

1bb06a11aef0925a0760dd4331e5a9bb74ed447f

-------

<p> Ban @Module(subcomponents) from DaggerAdapter until we properly support it

Supporting it correctly requires a fair amount of thought - Guice and Dagger scopes are different, multibindings are different, and implementing subcomponent factories+builders would be non-trivial. (We could possibly take advantage of child injectors to help?) Anyway, there's more to considered before traversing down this path.

fd95e8fa9cbb8f76f3669111fad6eb82c9c299a7

-------

<p> update the opensource asm to 7.1 & cglib to 3.2.12.

25979192a9f02e6da612fcdcafb68e4332a51103

-------

<p> Validate that all modules passed to DaggerAdapter are in fact Dagger modules

16e647d5e7e3252249e884f565f3199542731ee7

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,8 @@ cache:
 
 jdk:
   - oraclejdk11
-  - oraclejdk9
-  - oraclejdk8
+  - openjdk9
+  - openjdk8
 
 env:
   global:
--- a/core/src/com/google/inject/internal/ProviderMethod.java
+++ b/core/src/com/google/inject/internal/ProviderMethod.java
@@ -64,7 +64,10 @@ public abstract class ProviderMethod<T> extends InternalProviderInstanceBindingI
       Annotation annotation) {
     int modifiers = method.getModifiers();
     /*if[AOP]*/
-    if (!skipFastClassGeneration) {
+    if (!skipFastClassGeneration
+        // Protect against a bug in cglib where static interface methods were invoked with
+        // invokeinterface and not invokestatic. See https://github.com/cglib/cglib/pull/153
+        && instance != null) {
       try {
         net.sf.cglib.reflect.FastClass fc = BytecodeGen.newFastClassForMember(method);
         if (fc != null) {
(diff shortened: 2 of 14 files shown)


guillaumecle:
This pull request contained the upgrade of asm to 7.1 which is necessary when running on java 13. Is there a plan to release soon?

GedMarc:
@guillaumecle it looks like this repo has stalled -

GuicedEE has kept up to date, you can grab the 4.2.2 jre 13 here
https://search.maven.org/artifact/com.guicedee.services/guice 

The version updates with the encapsulating project, but it is fully in sync

Most of the extensions are available here - (The ones where the implementing library is also JMS compatible)
com.guicedee.services.extensions

PR #1251, opened 2019-06-17
Moe Sync

ronshapiro:

Commits:
=====
<p> Make @RequestScoped use @Scope instead of @ScopeAnnotation.

06256ad3abe926df07ec906084e60a3191873a93


ronshapiro:
@sameb it looks like the `ant` build is failing. Do we still use that? If so, it appears that we may need to install it ourselves (perhaps it's no longer installed?)

https://ant.apache.org/bindownload.cgi

sameb:
Yah looks like Travis stopped supplying it by default. I'll try to fix
tomorrow.

PR #1253, opened 2019-06-26
Update .travis.yml

ghatwala:
adding power support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,7 @@
 language: java
+os:
+  - linux
+  - linux-ppc64le
 sudo: false
 
 cache:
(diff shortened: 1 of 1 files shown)


sameb:
Can you elaborate on why you want to do this and what purpose it serves?

This will not change anything about Guice, it will just change the way it's continuous integration tests run.

ghatwala:
@sameb - we would want to have CI run to happen on different arch (ppc64le) too to have regression happening on power too fo this repo.

sameb:
Why?  Guice has no interaction with low level native architectures.  Do you have some reason to suspect otherwise?

ghatwala:
I signed it!

PR #1294, opened 2020-02-23
Simplified Test Syntax

Vince205:


sameb:
many of these things are the way it is precisely to test various java8 syntax.  changing most would defeat the purpose of the tests.

PR #1307, opened 2020-03-25
MOE Sync 2020-03-25

nglorioso:
This code has been reviewed and submitted internally. Feel free to discuss on
the PR, and we can submit follow-up changes as necessary.

Commits:
=====
<p> Adding Predicate<Method> to DaggerAdapter. Only the methods which satisfy the predicate will be processed by the DaggerAdapter. The objective is to conditionally limit Guice bindings exposed by the DaggerAdapter based on an external condition.

2d7a37b5918ff0d8cfbebc0ec3665f60a6a82cd5


nglorioso:
Covered by #1306

Issue #1193, opened 2018-06-21
Add support for Nullable as a TYPE_USE annotation


tbroyer:
I wanted to switch from JSR305 to Checker Framework `@Nullable`, but it breaks my `@Nullable` injection points:

```
1) null returned by binding at […]
 but the 1st parameter of […].<init>([…]) is not @Nullable
```

Apparently, this is because Guice only looks at field and parameter annotations, not TYPE_USE annotations: https://github.com/google/guice/blob/eface212cdf3eeb1579d6e033cd54fd1426d65ba/core/src/com/google/inject/spi/InjectionPoint.java#L117-L119

Note: I couldn't use the Java 7-compatible annotations from the Checker Framework either because it's named `@NullableDecl`, and Guice only looks at annotations named `@Nullable`.

tbroyer:
Possible workaround: in our case, I could easily switch to `OptionalBinder` and `Optional<T>` instead of `@Nullable T`.

ronshapiro:
FWIW, we decided (both Guice and Dagger) to ignore type annotations for qualifiers at least for the time being because of the complications that they may introduce into the key/type system that weren't previously considered.

This may be entirely different, as I'm not familiar with this part of Guice, but there's always the consideration of something like this:

```java
class Foo<T> {
  @Inject Foo(T t) {}
}

class Bar {
  @Inject Bar(Foo<@Nullable String> fooOfNullableString) {} 
}
```

nresare:
To clarify what I believe is happening here, the annotations that describe annotations and the reflection API to get information about annotations was extended in jdk8. Annotations such as `@Nullable` can now have a `@Target({TYPE_USE})` and this does indeed make sense to use for something like `@Nullable`. This is what for example he [checkerframework Nullable](https://checkerframework.org/api/org/checkerframework/checker/nullness/qual/Nullable.html) uses. To extract such annotation from a constructor, one need to use `constructor.getAnnotatedParameterTypes()...getAnnotations()` which was introduced in java8.

cpovirk:
There are a lot of things that "Guice supports `@Nullable` type annotations" could mean, but https://github.com/google/guice/commit/1a410a8bef1ec721e16f9f16a447235f09951f05 accomplished the main one. I assume there hasn't been a release yet, though.

sameb:
This was fixed in https://github.com/google/guice/commit/1a410a8bef1ec721e16f9f16a447235f09951f05.

PR #1311, opened 2020-04-09
Support @Nullable annoations marked with @Target({TYPE_USE})

nresare:
Java support for annotation metadata has evolved, and projects providing
implementations of @Nullable has been adopting those new features. This
has had the side effect of some commonly used @Nullable annotations not
working as expected with Guice. This change uses reflection to extract
new style @Nullable annotations as well as old style ones. Addresses
the issue described in https://github.com/google/guice/issues/1193


nresare:
@googlebot I signed it!

cpovirk:
https://github.com/google/guice/commit/1a410a8bef1ec721e16f9f16a447235f09951f05 has done some work for type annotations. That may be sufficient for your needs (though I don't believe there's been a release yet).

sameb:
Indeed that commit already fixed this.

PR #1313, opened 2020-04-17
Fix toString implementation of annotations for Java 14

### LINKED ISSUE #1312: The toString implementation of annotations has changed in Java 14
The explicitly-not-specified `toString` implementation of annotations has changed in Java 14. See JDK-8225356 - "Make javac's toString() on annotation objects consistent with core reflection" [1], for specific details.

At a minimum, annotations with just one element named `value` elide the "value=" prefix. This has some impact on the `toString` implementations in Guice. To be consistent with the Java 14 string representation of such annotations the "value=" should be dropped. This can be determined at runtime, so that behaviour can be consistent with the actual Java runtime behaviour.

For example:
```
diff --git a/core/src/com/google/inject/internal/Annotations.java b/core/src/com/google/inject/internal/Annotations.java
index a02b1d5d..1ffb6e53 100644
--- a/core/src/com/google/inject/internal/Annotations.java
+++ b/core/src/com/google/inject/internal/Annotations.java
@@ -221,6 +221,16 @@ public class Annotations {
     return QUOTE_MEMBER_VALUES ? "\"" + value + "\"" : value;
   }
 
+  private static final boolean SINGLE_VALUE_MEMBER_PREFIX = determineWhetherToElideValuePrefix();
+
+  /**
+   * Returns the member value key prefix {@code value=}, or the empty string. In Java
+   * 14, annotations with just one element named {@code value} elide the `value=` prefix.
+   */
+  public static String memberKeyString() {
+    return SINGLE_VALUE_MEMBER_PREFIX ? "value=" : "";
+  }
+
   @Retention(RUNTIME)
   private @interface TestAnnotation {
     String value();
@@ -240,6 +250,20 @@ public class Annotations {
     }
   }
 
+  @TestAnnotation("determineWhetherToElideValuePrefix")
+  private static boolean determineWhetherToElideValuePrefix() {
+    try {
+      String annotation =
+          Annotations.class
+              .getDeclaredMethod("determineWhetherToElideValuePrefix")
+              .getAnnotation(TestAnnotation.class)
+              .toString();
+      return annotation.contains("value=");
+    } catch (NoSuchMethodException e) {
+      throw new AssertionError(e);
+    }
+  }
+
   /** Checks for the presence of annotations. Caches results because Android doesn't. */
   static class AnnotationChecker {
     private final Collection<Class<? extends Annotation>> annotationTypes;
diff --git a/core/src/com/google/inject/internal/UniqueAnnotations.java b/core/src/com/google/inject/internal/UniqueAnnotations.java
index 36b213d2..2cb77898 100644
--- a/core/src/com/google/inject/internal/UniqueAnnotations.java
+++ b/core/src/com/google/inject/internal/UniqueAnnotations.java
@@ -51,7 +51,10 @@ public class UniqueAnnotations {
 
       @Override
       public String toString() {
-        return "@" + Internal.class.getName() + "(value=" + value + ")";
+        return String.format("@%s(%s%s)",
+                Internal.class.getName(),
+                Annotations.memberKeyString(),
+                value);
       }
 
       @Override
diff --git a/core/src/com/google/inject/name/NamedImpl.java b/core/src/com/google/inject/name/NamedImpl.java
index 9c4d10e4..dccc9784 100644
--- a/core/src/com/google/inject/name/NamedImpl.java
+++ b/core/src/com/google/inject/name/NamedImpl.java
@@ -53,7 +53,10 @@ class NamedImpl implements Named, Serializable {
 
   @Override
   public String toString() {
-    return "@" + Named.class.getName() + "(value=" + Annotations.memberValueString(value) + ")";
+    return String.format("@%s(%s%s)",
+            Named.class.getName(),
+            Annotations.memberKeyString(),
+            Annotations.memberValueString(value));
   }
 
   @Override
```

Tests can then be updated similar to:
   ```
$ git diff core/test/com/google/inject/MembersInjectorTest.java
diff --git a/core/test/com/google/inject/MembersInjectorTest.java b/core/test/com/google/inject/MembersInjectorTest.java
index 17869cd8..039d09f5 100644
--- a/core/test/com/google/inject/MembersInjectorTest.java
+++ b/core/test/com/google/inject/MembersInjectorTest.java
@@ -280,7 +280,7 @@ public class MembersInjectorTest extends TestCase {
       assertContains(
           expected.getMessage(),
           "1) No implementation for com.google.inject.MembersInjector<java.lang.String> "
-              + "annotated with @com.google.inject.name.Named(value="
+              + "annotated with @com.google.inject.name.Named(" + Annotations.memberKeyString()
               + Annotations.memberValueString("foo")
               + ") was bound.");
     }
```

[1] https://bugs.openjdk.java.net/browse/JDK-8225356

ChrisHegarty:
Fixes #1312

The explicitly-not-specified toString implementation of annotations has changed in Java 14. See JDK-8225356 - "Make javac's toString() on annotation objects consistent with core reflection" [1], for specific details.

At a minimum, annotations with just one element named value elide the "value=" prefix. This has some impact on the toString implementations in Guice. To be consistent with the Java 14 string representation of such annotations the "value=" should be dropped. This can be determined at runtime, so that behaviour can be consistent with the actual Java runtime behaviour.

[1] https://bugs.openjdk.java.net/browse/JDK-8225356

### DIFF
--- a/core/src/com/google/inject/internal/Annotations.java
+++ b/core/src/com/google/inject/internal/Annotations.java
@@ -221,14 +221,14 @@ public class Annotations {
     return QUOTE_MEMBER_VALUES ? "\"" + value + "\"" : value;
   }
 
-  private static final boolean SINGLE_MEMBER_PREFIX = determineWhetherToElideKePrefix();
+  private static final boolean SINGLE_VALUE_MEMBER_PREFIX = determineWhetherToElideValuePrefix();
 
   /**
    * Returns the member value key prefix {@code value=}, or the empty string. In Java
    * 14, annotations with just one element named {@code value} elide the `value=` prefix.
    */
   public static String memberKeyString() {
-    return SINGLE_MEMBER_PREFIX ? "value=" : "";
+    return SINGLE_VALUE_MEMBER_PREFIX ? "value=" : "";
   }
 
   @Retention(RUNTIME)
@@ -250,12 +250,12 @@ public class Annotations {
     }
   }
 
-  @TestAnnotation("determineWhetherToElideKePrefix")
-  private static boolean determineWhetherToElideKePrefix() {
+  @TestAnnotation("determineWhetherToElideValuePrefix")
+  private static boolean determineWhetherToElideValuePrefix() {
     try {
       String annotation =
           Annotations.class
-              .getDeclaredMethod("determineWhetherToElideKePrefix")
+              .getDeclaredMethod("determineWhetherToElideValuePrefix")
               .getAnnotation(TestAnnotation.class)
               .toString();
       return annotation.contains("value=");
--- a/core/src/com/google/inject/internal/Annotations.java
+++ b/core/src/com/google/inject/internal/Annotations.java
@@ -221,6 +221,16 @@ public class Annotations {
     return QUOTE_MEMBER_VALUES ? "\"" + value + "\"" : value;
   }
 
+  private static final boolean SINGLE_MEMBER_PREFIX = determineWhetherToElideKePrefix();
+
+  /**
+   * Returns the member value key prefix {@code value=}, or the empty string. In Java
+   * 14, annotations with just one element named {@code value} elide the `value=` prefix.
+   */
+  public static String memberKeyString() {
+    return SINGLE_MEMBER_PREFIX ? "value=" : "";
+  }
+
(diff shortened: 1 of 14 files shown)


sameb:
This has since been fixed.

Issue #251, opened 2014-07-07
@ProvidedBy @Singleton doesn't work


gissuebot:
_From [limpbizkit](https://code.google.com/u/limpbizkit/) on September 26, 2008 13:12:47_

It should provide a single instance.

The rules for `@`ProvidedBy and annotations don't need to be the same as for
`@`ImplementedBy, where scoping conflicts could be problematic. For
`@`ProvidedBy the scopes don't conflict.

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=251_

birdmanmandbir:
> _From [Ben.Manes](https://code.google.com/u/116561040243839055672/) on September 17, 2013 14:00:22_
> 
> Any chance this is fixed in Guice 4.0? Just ran into this for a servlet filter.

It is still not fixed. I think it is because `@Providedby` have larger priority than `@Singleton`?

GedMarc:
isn't it an ordering thing?
```@ProvidedBy``` is a lazy load provider, singletons with providers need to be registered during module creation as a new provided class?
If you are going eager singleton even more so...

birdmanmandbir:
> isn't it an ordering thing?
> `@ProvidedBy` is a lazy load provider, singletons with providers need to be registered during module creation as a new provided class?
> If you are going eager singleton even more so...

Do you mean if we use both `@Providedby` and `@Singleton`, there will be two providers, a lazy one and an eager one? 

Maybe singletons with providers is executed at the beginning, and `@ProvidedBy` is executed later, and ProvidedBy's result overwrite the previous result?

GedMarc:
No I mean there are two types are singletons, - 
But either way the outcome is the same, so you can't use ```@ProvidedBy``` for singletons.. 
this seems a very by design item.

birdmanmandbir:
> No I mean there are two types are singletons, -
> But either way the outcome is the same, so you can't use `@ProvidedBy` for singletons..
> this seems a very by design item.

When we use `@ProvidedBy`, `bind(Class1.class).in(Scopes.SINGLETON);` is required to get singleton
When we use `@Singleton`, guice will create a singleton provider to give a singleton, 
So if we both have `@ProvidedBy` and `@Singleton`, only `@ProvidedBy` is used by guice so we can't get singleton without `).in(Scopes.SINGLETON);`

Am I understanding it correctly?

And I also write more test cases:
[...]

PR #1319, opened 2020-05-03
Fix bug (#251)

birdmanmandbir:
Solve issue #251 and add corresponding test for it
Scope annotation in ProvidedBy handling should be added before create provided by binding

### DIFF
--- a/core/test/com/google/inject/ProvidedByAndScopeTest.java
+++ b/core/test/com/google/inject/ProvidedByAndScopeTest.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright (C) 2006 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


core/test/com/google/inject/ProvidedByAndScopeTest.java
```diff
@@ -0,0 +1,69 @@
...
+    public MixClass get() {
+      return new MixClass();
+    }
+  }
+
+  @Singleton
+  @ProvidedBy(MixClassProvider.class)
+  static class MixClass {
+    @Override
+    public boolean equals(Object obj) {
```
ogregoire, line 36: This method is useless as it's the default behavior of Object.

core/test/com/google/inject/ProvidedByAndScopeTest.java
```diff
@@ -0,0 +1,69 @@
...
+      return this == obj;
+    }
+  }
+
+  @Test
+  public void testMixClass() throws Exception {
+    final Injector injector =
+        Guice.createInjector(
+            new AbstractModule() {
```
ogregoire, line 45: You can just call `Injector injector = Guice.createInjector();` without any module for the same value.

core/test/com/google/inject/ProvidedByAndScopeTest.java
birdmanmandbir: OK，I will update it today : ) 
Thanks for your advice!

core/test/com/google/inject/ProvidedByAndScopeTest.java
birdmanmandbir: OK，I will update it today : ) 
Thanks for your advice!

PR #1322, opened 2020-05-06
FactoryProvider2: Remove illegal relfective access

### LINKED ISSUE #1321: WARNING: Illegal reflective access by com.google.inject.assistedinject.FactoryProvider2
I'm seeing this warning when starting Gerrit with Java11:

```
$ java -jar bazel-bin/release.war init -d ../test_site_gerrit_3.2 --batch --dev
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.assistedinject.FactoryProvider2 (file:/home/davido/.gerritcodereview/tmp/gerrit_6966619352392581698_app/guice-assistedinject-4.2.3.jar) to constructor java.lang.invoke.MethodHandles$Lookup(java.lang.Class,int)
WARNING: Please consider reporting this to the maintainers of com.google.inject.assistedinject.FactoryProvider2
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

The code has even "this isn't a public API" comment:

https://github.com/google/guice/blob/master/extensions/assistedinject/src/com/google/inject/assistedinject/FactoryProvider2.java#L900-L915

```java
// Note: this isn't a public API, but we need to use it in order to call default methods on (or
  // with) non-public types.  If it doesn't exist, the code falls back to a less precise check.
  private static final Constructor<MethodHandles.Lookup> methodHandlesLookupCxtor =
      findMethodHandlesLookupCxtor();

  private static Constructor<MethodHandles.Lookup> findMethodHandlesLookupCxtor() {
    try {
      Constructor<MethodHandles.Lookup> cxtor =
          MethodHandles.Lookup.class.getDeclaredConstructor(Class.class, int.class);
      cxtor.setAccessible(true);
      return cxtor;
    } catch (ReflectiveOperationException ignored) {
      // Ignore, the code falls back to a less-precise check if we can't create method handles.
      return null;
    }
  }
```

See also this upstream issue: [1].

[1] https://crbug.com/gerrit/12639

davido:
Fixes #1321.

The disabled test is referencing #904 that was reported by me for Gerrit
project. However, with this change, Gerrit works as expected.

### DIFF
--- a/extensions/assistedinject/src/com/google/inject/assistedinject/FactoryProvider2.java
+++ b/extensions/assistedinject/src/com/google/inject/assistedinject/FactoryProvider2.java
@@ -899,9 +899,10 @@ final class FactoryProvider2<F>
 
   // Note: this isn't a public API, but we need to use it in order to call default methods on (or
   // with) non-public types.  If it doesn't exist, the code falls back to a less precise check.
-  private static final Constructor<MethodHandles.Lookup> methodHandlesLookupCxtor =
-      findMethodHandlesLookupCxtor();
+  private static final Constructor<MethodHandles.Lookup> methodHandlesLookupCxtor = null;
+      //findMethodHandlesLookupCxtor();
 
+/*
   private static Constructor<MethodHandles.Lookup> findMethodHandlesLookupCxtor() {
     try {
       Constructor<MethodHandles.Lookup> cxtor =
@@ -913,6 +914,7 @@ final class FactoryProvider2<F>
       return null;
     }
   }
+*/
 
   private static MethodHandle createMethodHandle(Method method, Object proxy) {
     if (methodHandlesLookupCxtor == null) {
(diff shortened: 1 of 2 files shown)


kibertoad:
This doesn't look to be a correct solution.

extensions/assistedinject/src/com/google/inject/assistedinject/FactoryProvider2.java
```diff
@@ -899,9 +899,10 @@ protected Object initialValue() {
 
   // Note: this isn't a public API, but we need to use it in order to call default methods on (or
   // with) non-public types.  If it doesn't exist, the code falls back to a less precise check.
-  private static final Constructor<MethodHandles.Lookup> methodHandlesLookupCxtor =
-      findMethodHandlesLookupCxtor();
+  private static final Constructor<MethodHandles.Lookup> methodHandlesLookupCxtor = null;
+      //findMethodHandlesLookupCxtor();
```
kibertoad, line 903: Are you sure you were planning to commit all the commented out code?

HugoDeLugo:
Hi @davido,

could you please review the comment of @kibertoad?

I really need this fix. In this moment, I can't work with factories and I have to.

Thank you!

davido:
Fixed with different approach.

Issue #216, opened 2014-07-07
provider methods can't bind eager singletons


gissuebot:
_From [medotin](https://code.google.com/u/108745527391139638464/) on June 25, 2008 09:39:59_

We need an `@`Eager or something for provider methods.

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=216_

kuc:
Any chance that this will get ever fixed?

yaitskov:
use spring ;)

ogregoire:
Any news on this topic?

GedMarc:
```
  bind(Integer.class).toProvider(() -> -1).asEagerSingleton();
 System.out.println("Integer retrieved = " + get(Integer.class));
```
4.2.2 - works perfectly.

ogregoire:
@GedMarc  Thanks for the lovely words... Anything constructive you want to add? Your first post added maybe something valuable to people who might not know alternatives. But the second? I'm not so sure...

GedMarc:
@ogregoire Don't dish it out if you can't have it served. You gave sass, you got sass. You asked a question, I answered it.

Extract from the wiki :
https://github.com/google/guice/wiki/ProviderBindings

In essence, update to version 4, and use version 4 patterns. Don't use @Provider in modules, create Provider classes or use static inner anonymous (lambda's etc) which is the correct way.

https://stackoverflow.com/questions/22147817/binding-provides-method-as-eager-singleton

ogregoire:
@GedMarc More lovely words...

You didn't answer the question: are you working in the Guice team? Do you have internal news about this feature request you want to share? If you can't answer yes to any of those two questions, you probably should have thought about those before trying to provide an answer to my question. So, thanks for your input whatever your input is. I'll leave it as it.

Now if anyone in charge of the project could spend one minute providing an acceptable answer, I would pretty much appreciate. Is this idea behind this issue still valuable for you in the long term? Can anyone, including me to my best effort, provide a PR to solve this issue?

sameb:
Folks, please be kind to each other. 

As for the question at hand, there's no built-in to do what you want exactly. You can workaround the problem by introducing an @EagerSingleton annotation, having it delegate to Singleton scope, and manually iterating through bindings in the injector and eagerly instantiating bindings that are eager singletons.

sameb:
[...]

PR #1325, opened 2020-05-10
The enhancement for the issue #216

ShaneXiangH:
Solve issue #216 and add corresponding test for it.

I add an new way to create an eager singleton: use the annotation @EagerSingleton. 

In this way, when we do development in Stage.DEVELOPMENT, we can simply use the annotation to create an eager Injector, just like that in Stage.PRODUCTION and  use @Singleton to create a lazy injector.

**Test scenario**
When user both use @EagerSingleton in production stage, the injector should give an eager singleton instance.

When user both use @EagerSingleton in development stage, the injector should also give an eager singleton instance.

### DIFF
--- a/core/src/com/google/inject/EagerSingleton.java
+++ b/core/src/com/google/inject/EagerSingleton.java
@@ -1,0 +1,19 @@
+package com.google.inject;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.Target;
+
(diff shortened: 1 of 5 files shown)


ShaneXiangH:
@googlebot I signed it!

sameb:
Hello, thank you the PR. Without getting into the merits of whether or not @EagerSingleton is a good idea... the PR as-is appears to have other cruft in it (`isGuiceEnhanced`, etc..). I'm going to close this PR as it's old and not submittable. I'll shortly be commenting on the original issue.

PR #1337, opened 2020-05-30
fix bug in issue 973 - check linked key in jitbinding before inject with parent

birdmanmandbir:
Fix issue 973 and add test for it. #973 

Bug in issue 973 is happened because not inject correct 
injector into the instance.
It can be solved by just check whether child injector 
already has linked key before use parent to


birdmanmandbir:
Not sure why the CI not passed because 404 not found

deinlandel:
Hello. Any ETA on this getting into release? We are affected by this bug.

sameb:
While I sympathize with the issue and the PR (and agree that the binding API leads the user down the wrong path here), this unfortunately is not submittable. 

If we force linked bindings to stay in the child, then this can materially alter graphs such that:
  * Two sibling children will stop sharing an implementation in a scenario of (child1: bind Foo to Impl, child2: bind Foo to Impl]. You could argue that they never _intended_ to share an impl... but we can't know that ahead of time.
  *  A parent injector may later need to access the RHS of the binding, which it can do today (if not explicitly bound in the child). If we change that, the parent injector's usage will break.

I'll instead submit some documentation updates warning of this.

Issue #1236, opened 2019-03-05
guice-grapher extension does not graph bindings in lower injector's modules


Octogonapus:
The `guice-grapher` extension can graph the bindings for a child injector fine but does not graph any bindings for the injector the child injector was created from. To illustrate this point, look at the following code to reproduce the problem:
```kotlin
import com.google.inject.Guice
import com.neuronrobotics.bowlerbuilder.view.cad.cadengine.util.GuiceGrapher
import org.jlleitschuh.guice.module
import org.junit.jupiter.api.Test

interface TestInterfaceA
class TestClassA : TestInterfaceA

interface TestInterfaceB
class TestClassB : TestInterfaceB

internal class TestGrapher {

    @Test
    fun `test grapher`() {
        val injectorA = Guice.createInjector(module {
            bind<TestInterfaceA>().to<TestClassA>()
        })

        val injectorB = injectorA.createChildInjector(module {
            bind<TestInterfaceB>().to<TestClassB>()
        })

        GuiceGrapher.graph("test.dot", injectorB)
    }
}
```

The graph this produces is:
![image](https://user-images.githubusercontent.com/4064722/53776819-49413380-3ec5-11e9-8371-62de1ab0b26b.png)

The code for `GuiceGrapher` is:
```java
public class GuiceGrapher {

  public static void graph(final String filename, final Injector inj) throws Exception {
    final ByteArrayOutputStream baos = new ByteArrayOutputStream();
    PrintWriter out = new PrintWriter(baos);

    final Injector injector = Guice.createInjector(new GraphvizModule());
    final GraphvizGrapher renderer = injector.getInstance(GraphvizGrapher.class);
    renderer.setOut(out);
    renderer.setRankdir("TB");
    renderer.graph(inj);

    out = new PrintWriter(new File(filename), "UTF-8");
    String s = baos.toString("UTF-8");
    s = fixGrapherBug(s);
    s = hideClassPaths(s);
    out.write(s);
    out.close();
  }

  private static String hideClassPaths(String s) {
    s = s.replaceAll("\\w[a-z\\d_\\.]+\\.([A-Z][A-Za-z\\d_\\$]*)", "$1");
    s = s.replaceAll("value=[\\w-]+", "random");
    return s;
  }

  private static String fixGrapherBug(String s) {
    s = s.replaceAll("style=invis", "style=solid");
    s = s.replaceAll(" margin=(\\S+), ", " margin=\"$1\", ");
    return s;
  }
```

PR #1339, opened 2020-05-31
Fixed issue #1236

CHENGWEI-inf:
We tried to call parent when create bindings in Grapher


birdmanmandbir:
add a link to issue: #1236

sameb:
This does indeed look like a problem with the Grapher, but the PR is not submittable. If you'd like to submit the fix, please:

 1) Sign the google CLA
 2) Improve the PR so that it graphs _all_ the parent bindings, not just the first two.
 3) Support chains of >1 parent injector.
 4) Add a test.

Since it's been a while since this was opened, I'm going to close the PR. But please feel free to reopen a new one with the changes. Thanks!

PR #1346, opened 2020-06-09
Actually call default methods in AssistedInject factories

### LINKED ISSUE #1345: AssistedInject and default methods
I have a factory interface, extended by factory subtypes. The factories have default methods. So I would expect that they are called instead of whatever implementation is used.

Here's my full test case. Yes, I'm using a small hierarchy because that's actually part of my actual needs. It's probably not needed to showcase the error, though.

    class DefaultMethodTest {

	static class Foo { String str; @Inject Foo(@Assisted String str) { this.str = str; } }
	static class Bar extends Foo { @Inject Bar(@Assisted String str) { super(str); } }
	
	interface FooFactory {
		Foo create(@Assisted String str);
		default String id() { return "Foo"; }
	}
	interface BarFactory {
		Bar create(@Assisted String str);
		default String id() { return "Bar"; }
	}
		
	@Test
	void testAssistedDefault() {
		var injector = Guice.createInjector(binder -> binder.install(new FactoryModuleBuilder().build(BarFactory.class)));
		var barFactory = injector.getInstance(BarFactory.class);
		assertThat(barFactory.id()).isEqualTo("Bar");
	}
    }

The assertion fails because the call returns the empty string `""`. It must be a default or something like that, I'm not entirely sure, but I would have expected that my default method is called.

ogregoire:
Fixes #1345


ogregoire:
Sorry, I did commit with a wrong e-mail address. Reopening in another, correct repo.

PR #1349, opened 2020-06-14
Fix #716: requireExplicitBindings support requiring explicit binding specific classes

### LINKED ISSUE #716: requireExplicitBindings() method should support requiring explicit binding on individual classes
_From [glenviewjeff](https://code.google.com/u/110087215095127878251/) on July 19, 2012 16:32:35_

I'm using a couple of different injectors and I want to ensure I don't accidentally access a class that's bound automatically from an injector, especially since its scope is important.

For instance if I have a method Foo that is "illegally" requesting a Bar:

class Bar { `@`Inject Bar() {} } // Shouldn't be accessible in Foo's injector
class Foo { `@`Inject Foo(Bar bar) {} } // Binding should fail here

Apparently, the only way to do this presently is to force a run-time error with a provider method.  

I'd like Guice support to force an automatic binding to fail only on a specific class.&nbsp;&nbsp;I think it could work via requireExplicitBindings() with variable length parameter list of classes to prohibit.

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=716_

akhomchenko:
### Description

As per title. I have a real life example which made me search for this issue:

`ClientConfiguration` from AWS SDK is shared a lot in a project that I work on. It has a default constructor which makes it easy to accidentally require default implementation. I've created a custom qualifier to make sure that `ClientConfiguration` has an explicit binding. I think it is not the best solution and so I ended up writing this.

### Notes

* not so many tests now. I'm glad to add more if you would agree to have this feature as part of Guice.
* I'm not very good at benchmarks but I've tried to test field vs method access (as I have replaced direct `.jitDisabled` call with `.isJitDisabled(...` method) with JMH and haven't spotted a significant difference for a common case (`Predicates#alwaysTrue()` is invoked). Of course there is a difference when `requireExplicitBindings` is invoked with parameters.

### DIFF
--- a/core/test/com/google/inject/JitBindingsTest.java
+++ b/core/test/com/google/inject/JitBindingsTest.java
@@ -19,6 +19,7 @@ package com.google.inject;
 import static com.google.common.collect.ImmutableSet.of;
 import static com.google.inject.Asserts.assertContains;
 import static com.google.inject.JitBindingsTest.GetBindingCheck.ALLOW_BINDING;
+import static com.google.inject.JitBindingsTest.GetBindingCheck.ALLOW_BINDING_PROVIDER;
 import static com.google.inject.JitBindingsTest.GetBindingCheck.FAIL_ALL;
 
 import java.util.Set;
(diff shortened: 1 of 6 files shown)


core/test/com/google/inject/JitBindingsTest.java
```diff
@@ -390,6 +440,82 @@ protected void configure() {
...
+    Injector child =
+        parent.createChildInjector(new AbstractModule() {
+          @Override
+          protected void configure() {
+            binder().requireExplicitBindings(Bar.class);
+          }
+        });
+    ensureFails(child, FAIL_ALL, Foo.class, FooImpl.class);
+    // Unfortunately Bar might be injected via Binding if it was JITed previously by parent
+    ensureFails(child, ALLOW_BINDING_PROVIDER, Bar.class);
```
gagoman, line 516: To be honest, I'm not sure what is better here, please advise.

akhomchenko:
Friendly ping. Will you consider this pull request?

sameb:
Thank you for this PR.  This is an interesting idea, but unfortunately I don't think the additional API is worthwhile to include. This is because of a few reasons:
  * `requireExplicitBindings` is intended as a wholesale measure to prevent accidental JIT usage of _anything_.  OTOH`, `requireExplicitBindings(Class...)` would be a targeted measure, and can be sidestepped by not including that module that requires explicit bindings for that class, or by including a `bind(ThatClass.class)` statement, which defeats the initial goal of "don't let people use the <this type>".
  *  Guice has since added a @RestrictedBindingSource annotation (see https://github.com/google/guice/wiki/RestrictedBindingSource), which feels like it's slightly better suited for your use-case of wanting to prevent people from using the "wrong" type. Unfortunately, @RestrictedBindingSource doesn't apply to your use-case, because it's intended as an annotation to prevent folks from _binding_ things, as opposed to preventing folks from _using_ things. It's also limited to restricted types you own, because the type or qualifier itself has to be annotated. However, I could imagine extending @RestrictedBindingSource to support some kind of Binder integration, like `binder.restrictBindingSource(RestrictionInfo, ClassOrAnnotation...)`, or something similar to that, which could potentially prevent usage + other bindings. This may not make sense, though, because folks could just bind the types by removing that module altogether.
  * Overall, this feels like something that's better suited to a test: use the [Guice SPI](https://github.com/google/guice/wiki/ExtendingGuice) to iterate over what's injected and confirm that no "bad" types are used.

Issue #785, opened 2014-07-07
Deadlock on "waitForValue"


gissuebot:
_From [SmokejumperIT](https://code.google.com/u/114670102270110551777/) on November 29, 2013 14:58:27_

This is using Guice 3.0. When we have multiple injections of the same class happening at the same time, I'm encountering a deadlock here:

&nbsp;&nbsp;&nbsp;java.lang.Thread.State: WAITING (on object monitor)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at java.lang.Object.wait(Native Method)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at java.lang.Object.wait(Object.java:503)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.internal.util.$MapMaker$StrategyImpl.waitForValue(MapMaker.java:529)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;- locked &lt;0x00000000ea6f4248&gt; (a com.google.inject.internal.util.$MapMaker$LinkedStrongEntry)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.internal.util.$MapMaker$StrategyImpl$FutureValueReference.waitForValue(MapMaker.java:619)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.internal.util.$MapMaker$StrategyImpl.waitForValue(MapMaker.java:533)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.internal.util.$MapMaker$StrategyImpl.waitForValue(MapMaker.java:419)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.internal.util.$CustomConcurrentHashMap$ComputingImpl.get(CustomConcurrentHashMap.java:2061)
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;at com.google.inject.internal.FailableCache.get(FailableCache.java:50)
[...]

PR #1389, opened 2020-09-01
Test that crashes Guice inside Guava

swankjesse:
This test causes Guice to violate the requirements of Guava's
LoadingCache by triggering a recursive load.

The recursive load crashes either with an UncheckedExecutionException
(bad) or causes a deadlock (terrible).

https://github.com/google/guice/issues/785


swankjesse:
PR that includes this commit plus another commit with a fix:
https://github.com/google/guice/pull/1394

PR #1394, opened 2020-09-03
Fix issues with recursive loading

swankjesse:
Crash with nice Guice errors rather than Guava crashes or deadlocks.

### DIFF
--- a/core/src/com/google/inject/internal/FailableCache.java
+++ b/core/src/com/google/inject/internal/FailableCache.java
@@ -16,65 +16,103 @@
 
 package com.google.inject.internal;
 
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
 import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Maps;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.FutureTask;
 
 /**
  * Lazily creates (and caches) values for keys. If creating the value fails (with errors), an
  * exception is thrown on retrieval.
  *
+ * This defends against re-entrant loads, and removals of elements while they are being loaded.
+ *
  * @author jessewilson@google.com (Jesse Wilson)
  */
 public abstract class FailableCache<K, V> {
 
-  private final LoadingCache<K, Object> delegate =
-      CacheBuilder.newBuilder()
-          .build(
-              new CacheLoader<K, Object>() {
-                @Override
-                public Object load(K key) {
-                  Errors errors = new Errors();
-                  V result = null;
-                  try {
-                    result = FailableCache.this.create(key, errors);
-                  } catch (ErrorsException e) {
-                    errors.merge(e.getErrors());
-                  }
-                  return errors.hasErrors() ? errors : result;
-                }
-              });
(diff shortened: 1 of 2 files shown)


core/src/com/google/inject/internal/FailableCache.java
```diff
@@ -16,65 +16,103 @@
...
-                  V result = null;
-                  try {
-                    result = FailableCache.this.create(key, errors);
-                  } catch (ErrorsException e) {
-                    errors.merge(e.getErrors());
-                  }
-                  return errors.hasErrors() ? errors : result;
-                }
-              });
+  private final Map<K, FutureTask<V>> delegate = new LinkedHashMap<>();
```
ad-fu, line 37: "curious why this needs to be a linked map instead of just a plain map.  if it doesn't need to be linked, we could get rid of explicit synchronization using ConcurrentHashMap instead AFAICT.

(only question of switching to concurrent would be atomicity of calling create(..) during the computeIfAbsent->newFutureTask.  unsure if atomic, unsure if needs to be.)"

core/src/com/google/inject/internal/FailableCache.java
```diff
@@ -16,65 +16,103 @@
...
+    try {
+      FutureTask<V> futureTask;
+      synchronized (delegate) {
+        futureTask = delegate.computeIfAbsent(key, this::newFutureTask);
+      }
+
+      futureTask.run();
+      result = futureTask.get();
+    } catch (InterruptedException e) {
+      throw new RuntimeException(e);
```
ad-fu, line 60: "add comments explaining why we rethrow here (and the non-ErrorsException instanceof below)?"

core/src/com/google/inject/internal/FailableCache.java
```diff
@@ -16,65 +16,103 @@
...
-        });
+    synchronized (delegate) {
+      ImmutableMap.Builder<K, V> result = ImmutableMap.builder();
+      for (Map.Entry<K, FutureTask<V>> entry : delegate.entrySet()) {
+        FutureTask<V> futureTask = entry.getValue();
+        try {
+          if (futureTask.isDone()) {
+            result.put(entry.getKey(), futureTask.get());
+          }
+        } catch (Exception ignored) {
```
ad-fu, line 112: "doc why its safe to ignore the exception -- namely, it was already handled above in get().  that said, it might be nice to switch this back to the old style where it returned V|Errors, which will get rid of a lot of exception handling.  it'll introduce some more casting, but IMO ditching the exception handling will end up with cleaner code even w/ the casting.  up to you."

ad-fu (commented):
Copied-across comments from the internal CL.

core/src/com/google/inject/internal/FailableCache.java
```diff
@@ -16,65 +16,103 @@
...
+    }
+
+    V result = null;
+    try {
+      FutureTask<V> futureTask;
+      synchronized (delegate) {
+        futureTask = delegate.computeIfAbsent(key, this::newFutureTask);
+      }
+
+      futureTask.run();
```
vlsi, line 57: It looks like multiple concurrent threads could attempt to run the same futuretask. Is it what you intended?
Thypically only one thread should run, and the rest wait on .get

swankjesse:
I’m reviewing old PRs and I abandoned this one before I pushed it over the finish line. That’s my fault.

As an alternative, please consider reviewing and merging #1389. That PR demonstrates the bug without a specified fix. Y’all can figure out what fix you like best!

Issue #1490, opened 2021-02-19
Support for jakarta annotations


hejcz:
Do you have any time estimations for supporting jakarta.* annotations? We want to migrate to Jersey 3 but it's impossible because `GuiceFilter` implements `javax.servlet.Filter`. Another problem is #1383.

kashike:
#1383

hejcz:
#1383 mentions only jakarta.inject so I don't know if it adresses the same scope of changes.

GedMarc:
You can also use 
https://search.maven.org/artifact/com.guicedee.services/guice

which is a compiled guice (from master branch/5) with all references on jakarta -

onacit:
@GedMarc Check [this](https://github.com/GedMarc/GuicedEE/issues/2) out, yo.

docent:
Bumping this for visibility. We have to recompile Guice in order to move to to the new jakarta.inject and jakarta.servlet packages. Not ideal...

krisso-rtb:
Logback switched from `javax.servlet-api` to `jakarta.servlet-api` in version `1.4.0`.
We have `guice 5.1.0` which (still) doesn't have support for `jakarta.servlet-api`.
It basically means **we're blocked** with upgrading to the newer logback.

Could we somehow fix the World? 😄

froque:
@krisso-rtb there is logback 1.3 to solve that problem https://logback.qos.ch/news.html

But your point remains, the lack of Jakarta support in Guice is a show stopper for us

ben-manes:
@sameb could we get something simple and low effort to be forward compatible? Maybe a new extension, `guice-jakarta-servlet`, that forks the old one with the updated imports? I don't think the community is asking for much, just a little support so that we can update our http servlet engines (tomcat, jetty, etc).

sameb:
I haven't followed the discussion... but I think you'll find that if we
bundled a different servlet extension, it would suffer from a lack of
maintenance. While basically ~every java server in Google depends on Guice
(and thus we're careful to keep it stable and working well), nothing would
depend on this new extension. Very few servers even depend on the existing
servlet extension (as we primarily use a different technology internally).

Other folks are in a better position to ensure a new servlet extension
exists and is maintained. If you're just looking for a fork of the servlet
extension, there shouldn't be anything special about the existing one that
requires co-bundling/shipping privileges.

OTOH, if you're looking for a fork of core Guice itself, that's unlikely to
happen.

ben-manes:
[...]

PR #1541, opened 2021-09-29
Upgrade hibernate version to support JPA 3.0

oswaldquek:
Sort of related to "jakarta" naming issues in https://github.com/google/guice/issues/1490 and https://github.com/google/guice/issues/1383. Upgrading the hibernate library - which uses the `jakarta.persistence` namespace - in the persistence extension seems to work fine on its own. 

This PR might not be accepted as it leaves a mix of `jakarta.persistence` and `javax.*` in the guice library but thought I'd give it a try. At least it was pretty low effort to do!

### DIFF
--- a/extensions/persist/pom.xml
+++ b/extensions/persist/pom.xml
@@ -15,10 +15,9 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
-      <artifactId>hibernate-jpa-2.0-api</artifactId>
-      <version>1.0.0.Final</version>
-      <scope>provided</scope>
+      <groupId>org.hibernate</groupId>
+      <artifactId>hibernate-core-jakarta</artifactId>
+      <version>5.5.7.Final</version>
     </dependency>
     <dependency>
       <groupId>javax.servlet</groupId>
@@ -35,7 +34,7 @@
     <dependency>
       <groupId>org.hibernate</groupId>
       <artifactId>hibernate-entitymanager</artifactId>
-      <version>4.1.7.Final</version>
+      <version>5.5.6.Final</version>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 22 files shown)


oswaldquek:
@googlebot I signed it!

GedMarc:
@oswaldquek There is a Jakarta namespace for Guice already right now - It is under the guicedee.com maven group - and supports JPMS fully up the chain (For JLink builds)

Guice - https://mvnrepository.com/artifact/com.guicedee.services/guice
Guice Persist  - https://mvnrepository.com/artifact/com.guicedee.services/guice-persist
Guice Servlet - https://mvnrepository.com/artifact/com.guicedee.services/guice-servlet

For HIbernate Pre-Version 6 with Jakarta and Guice support you can also use the hibernate-x libraries under the same maven group

oswaldquek:
@GedMarc Thanks - I saw that on the other issues, however I'm not sure how official guicedee is. It also appears to have a different licence, and only 2 stars and 1 fork.

GedMarc:
@oswaldquek Yep, But I do know, or strongly under the impression, Guice for the time being is on the javax. namespace for the foreseeable future :) Still sorting out AOP on JDK's above version 9  with the new strict encapsulation rules.

Hibernate itself also has a JRT issue which is being resolved as well - https://hibernate.atlassian.net/browse/HHH-14858?jql=order%20by%20created%20DESC / https://github.com/hibernate/hibernate-orm/pull/4265

But someone had to do it yes? All of these things had to be found, and resolved, for us to continue development and build native runtimes -  2 star and 1 fork or not xD 

@mcculls Do we know if there is any target to move Guice to Jakarta namespace, or is it currently just the AOP being worked on?

oswaldquek:
@GedMarc Of course, reliability isn't determined by stars or fork. I did try using https://guicedee.com/dbmodules.html, and it worked 😄  Unfortunately however my organisation is more concerned with using official libraries and with the relevant licensing.

PR #1549, opened 2021-11-11
fix ci's settings. lack of java version in publish

keyno63:
CI failed `publish` job, the reason is  lacking of Java version setting , maybe.
(a `matrix.*` value can be used every task. the value was set in `test` job, but don't set in `publish` job)

Add java version to `java-version param` in `publish` job. Java11 is stable version.


sameb:
It looks like this has since been already fixed. Thank you for the PR, though.

PR #1553, opened 2021-12-10
Add creation of jar with shaded guava and asm.

morgwai:
Use maven-shade-plugin for all shading jobs instead of jarjar-maven-plugin.
Remove jarjar profile and move all jar building to the main build section.


sameb:
WDYT of naming it "mostly-shaded"?  I'd prefer to avoid directly referencing asm & guava, in case something else ever gets added.

morgwai:
@sameb 
My goal is to make classifiers self-explanatory, so that users are not forced to navigate to `pom.xml` and search for relevant section in that not-so-short file each time they wonder which jar they should use in their project. The more deps we decide to shade and more jars we produce, the more important it gets.

Nevertheless, if you have a strong feeling about it, then I will change it the way you suggest: please let me know.

morgwai:
@sameb @mcculls friendly ping :)  Please let me know what you think about this. Thanks!

GedMarc:
@morgwai I was looking at something similar as an option to avoid the inconsistencies of versions - the primefaces guys have an interesting solution - shading only the class (x3) that is actually required - 

I think bundling the whole of guava is a bit much personally (but I am doing it), 
From what I can see only one class from guava is actually utilized - can you confirm this?

ASM is already a modular jar and doesn't need shading at all? It's used by quite a few libraries as well and is always updated before a JDK release (because JDK itself uses ASM..........) - so I don't think ASM should get shaded at all

morgwai:
Hi @GedMarc
I'm using `minimizeJar` option, which is supposed to do exactly the same: include only the classes actually used + their transitive dependencies.
This reduced the size by only about 30% though, so maybe there's still some area for improvement somehow: maybe `minimizeJar` is too conservative in some cases or cannot detect that some execution paths will never be taken by Guice.

Regarding actual Guava usage in Guice: `grep -r com.google.common core/src/ | cut -d : -f 2 | grep '^import' | grep -v 'import static' | sort | uniq -c | wc -l` returns 42.

Regarding ASM: to be honest I currently lack knowledge to discuss whether it should be shaded also or not. It was shaded before (with the jarjar profile), so I've included it also in the new jar to make transition as smooth as possible.
Generally Sam wrote that the less is needed to be shaded, the happier he will be, so if you have some insights why shading ASM is not necessary, it would be great to discuss this on the mailing list :)
In this PR I'd prefer to do just the first baby-step though ;-)

For context: all previous discussions I'm referring to took place in this thread on the mailing-list: https://groups.google.com/g/google-guice/c/0inCjcpWUwc

Cheers!

sameb:
Hello,

I'm not sure where this PR landed, and also the CLA isn't signed. As such, I'm going to close this PR.  If someone wants to bring up the idea of changing the what we shade, please open a new PR and we can try again.

Thank you.

PR #1569, opened 2022-01-24
Make PrivateModule#binder() non-final to allow binder customization in subclasses

### LINKED ISSUE #1525: Make PrivateModule#binder() non-final
Use case: I want to add convenience methods to `AbstractModule` and `PrivateModule` (for easier Kotlin interop), and the issue is that most of the methods in `*Module` are `protected`, so the only way to add convenience methods for me is to subclass `AbstractModule` and `PrivateModule`.

However, the sub-classes are purely technical (`abstract class KotlinModule : AbstractModule` and `abstract class KotlinPrivateModule : PrivateModule`), so I want to exclude them from "source stack" so the exceptions point to the actual binder use rather than to the common `KotlinModule` and `KotlinPrivateModule`.

Here's what I do with `AbstractModule`, and it more-or-less works:

```kotlin
abstract class KotlinModule : AbstractModule() {
    override fun binder() =
        super.binder().skipSources(KotlinModule::class.java) // <-- Here I configure binder to ignore KotlinModule class

    protected inline fun <reified T : Any> bind() = bind(typeLiteral<T>())
    protected inline fun <reified T : Any> bind(name: String) =
        bind<T>().annotatedWith(Names.named(name))
...
```

The good part is that all default `AbstractModule` methods like `bind(TypeLiteral<T> typeLiteral)` use `binder()` method, so they use the overridden `binder()`, so all bindings get my `skipSources` configuration.

Unfortunately, that does not work for `PrivateModule` since `binder()` is final there.

```kotlin
abstract class KotlinPrivateModule : PrivateModule() {
    override fun binder() = // <-- this does not compile since binder() is final in PrivateModule
        super.binder().skipSources(KotlinModule::class.java)

    protected inline fun <reified T : Any> bind() = bind(typeLiteral<T>())
...
```

----

An alternative option is to make default `bind(..)` methods public, so Kotlin/Scala extensions could be added without subclassing `AbstractModule`.

vlsi:
One of the cases for overriding `#binder()` would be calling `Binder#skipSources(...)`

fixes #1525

### DIFF
--- a/core/src/com/google/inject/PrivateModule.java
+++ b/core/src/com/google/inject/PrivateModule.java
@@ -137,7 +137,7 @@ public abstract class PrivateModule implements Module {
   // everything below is copied from AbstractModule
 
   /** Returns the current binder. */
-  protected final PrivateBinder binder() {
+  protected PrivateBinder binder() {
     checkState(binder != null, "The binder can only be used inside configure()");
     return binder;
   }


vlsi:
@ronshapiro , sorry for pinging you, however, would you please review and merge the change?

It is kind of sad to wait one year for 1 line change 😭

vlsi:
@cpovirk , could you please review and merge the fix?

In a nutshell, `AbstractModule.binder()` method is `protected` and `non-final`: https://github.com/google/guice/blob/c7a74a108b2596d5f1b69c6f3dcf32c27f3b5dab/core/src/com/google/inject/AbstractModule.java#L74

However, `PrivateModule.binder()` is `protected` and `final` at the same time.
I believe that was just a typo, and there's no practical meaning for making `PrivateModule.binder()` method final.

I listed more details in the linked https://github.com/google/guice/issues/1525#issue-923731861 issue.

Issue #1555, opened 2021-12-14
JpaPersistService does not allow nested UnitOfWork


edechter:
JpaPersistService's implementation of UnitOfWork::begin looks like it throws an exception if called within an existing UnitOfWork: https://github.com/google/guice/blob/e960b66d3d5931b9cb1aebd49e452e2c489a921e/extensions/persist/src/com/google/inject/persist/jpa/JpaPersistService.java#L75

But the interface JavaDoc says that this should be okay: https://github.com/google/guice/blob/e960b66d3d5931b9cb1aebd49e452e2c489a921e/extensions/persist/src/com/google/inject/persist/UnitOfWork.java#L42. 

Is this an error?

crizzis:
I just created a [PR](https://github.com/google/guice/pull/1609/checks) solving this issue and adjusting the logic behind nested transactions as well. 

The CI build is stuck at the CLA step, even though I have been added to the CLA by my organization. Kindly requesting a status update on the CLA processing for [Digital New Agency](https://digitalnewagency.com/).

PR #1606, opened 2022-02-22
JpaPersistService does not allow nested UnitOfWork #1555

crizzis:


extensions/persist/src/com/google/inject/persist/finder/Finder.java
```diff
@@ -46,5 +48,5 @@
    * subclass of {@code java.util.Collection}.
    */
   @SuppressWarnings({"rawtypes"}) // Unavoidable because class literal uses raw types.
-  Class<? extends Collection> returnAs() default Collection.class;
+      Class<?> returnAs() default Collection.class;
```
crizzis, line 51: Needed to relax the type constraint due to `ReturnType.ARRAY`

extensions/persist/src/com/google/inject/persist/jpa/JpaTransactionInterceptor.java
```diff
@@ -0,0 +1,106 @@
...
+          .anyMatch(ignored -> ignored.isAssignableFrom(e.getClass()));
+    }
+
+    private boolean noMatchInOnRollback(Exception e) {
+      return Arrays.stream(rollbackOn)
+          .noneMatch(rollbackable -> rollbackable.isAssignableFrom(e.getClass()));
+    }
+
+    // TODO(user): Cache this method's results.
+    private static Transactional readTransactionMetadata(MethodInvocation methodInvocation) {
```
crizzis, line 88: I was wondering if `javax.transaction.Transactional` was ever supposed to be supported here

PR #1609, opened 2022-02-23
JpaPersistService does not allow nested UnitOfWork #1555

crizzis:


sameb:
Thanks for this PR. I recognize there is indeed a problem with JpaPersistService in that it doesn't allow nested units of work... however, I can't merge this PR because it is _way_ too big. If you'd like to submit a more focused PR that attempts to solve the problem-at-hand without the extraneous refactorings, then please do so.  Thank you!

PR #1633, opened 2022-06-02
Prevent deadlock on "waitForValue"

PaulFridrick:
I tried a workaround to prevent the deadlock on "waitForValue" in the Guava Cache used for the ConstructorInjectorStore (issue https://github.com/google/guice/issues/785).

- I no longer have the unhelpful stacktrace " java.lang.IllegalStateException: Recursive load"
- In some cases, the IllegalStateException was nether thrown and the thread just hung : this case is not fixed with my workaround (I was not able to reproduce the problem with unit tests)
- When writing the new unit tests I noticed another issue which is likely related to the root cause of the deadlock : in the cases where the IllegalStateException would occur we are able to get some bindings which Guice should not be able to create (see the added class test)...

I don't think this workaround should be merged as-is, but I hope it will help people find the proper fix...

I will try to dig further if I find the time.

### DIFF
--- a/core/src/com/google/inject/internal/Errors.java
+++ b/core/src/com/google/inject/internal/Errors.java
@@ -499,6 +499,10 @@ public final class Errors implements Serializable {
         t);
   }
 
+  public Errors errorCacheAlreadyLoading(Object key) {
+    return addMessage(ErrorId.OTHER, "%s was already loading.", key);
+  }
+
   public Errors errorNotifyingTypeListener(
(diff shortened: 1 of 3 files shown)


PaulFridrick:
So I improved the workaround :
- we should no longer have the unhelpful stacktrace "java.lang.IllegalStateException: Recursive load"
- it should no longer be possible for the thread to hang
- the new tests are all green

Issue #1510, opened 2021-04-08
Cycle detection in a multi-thread environment leads to OutOfMemoryError


gvisco:
In some scenario the cycle detection leads the application to an OutOfMemoryError.
Specifically, this seems to happen if:
- At least one object *O* involved in a cyclic dependency is a Singleton
- *O* has not been initialized yet
- Multiple threads require the injection of *O* at the same time

The following test reproduces the issue (not always but often)
```java
import java.util.concurrent.CountDownLatch;

import org.junit.Test;

import com.google.inject.AbstractModule;
import com.google.inject.Guice;
import com.google.inject.Inject;
import com.google.inject.Injector;
import com.google.inject.Singleton;

public class GuiceCircularDependenciesIT extends AbstractModule {

    static interface Store {
        void sell();
    }
    
    static interface Customer {
        void buy();
    }
    
    static class StoreImpl implements Store {
        private final Customer customer;

        @Inject public StoreImpl(Customer customer) {
            this.customer = customer;
        }

        @Override
        public void sell() { }
    }

    static class CustomerImpl implements Customer {

        private Store store;

        @Inject public CustomerImpl(Store store) {
            this.store = store;
        }
        
        @Override
        public void buy() {
            store.sell();
        }
    }
    
    @Override
    protected void configure() {
        bind(Store.class).to(StoreImpl.class);
        bind(Customer.class).to(CustomerImpl.class).in(Singleton.class);
    }
    
    @Test
    public void testCircularDependencies() {
        for (int j = 0; j < 100; ++j) {
            Injector injector = Guice.createInjector(this);
            CountDownLatch latch = new CountDownLatch(8);
            for (int i = 0; i < 8; ++i) {
                Thread thread = new Thread(() -> {
                    try {
                        Customer instance = injector.getInstance(Customer.class);
                        instance.buy();
                    } catch (Exception e) {
                        System.out.println(e);
                    } finally {
                        latch.countDown();
                    }
                });
                thread.setDaemon(true);
                thread.start();
            }
            try {
                latch.await();
                System.out.println("DONE: " + j);
            } catch (InterruptedException e) {
            }
        }

    }
}
```
a sample output, when the issue occurs, is
```
DONE: 0
DONE: 1
DONE: 2
DONE: 3
DONE: 4
Exception in thread "Thread-47" java.lang.OutOfMemoryError: Java heap space
    at java.base/java.util.Arrays.copyOf(Arrays.java:3689)
[...]

PR #915, opened 2015-04-21
Merge internal changes

sameb:
Notably, this changes the singleton scope to get rid of the injector-wide lock.

PR #1635, opened 2022-06-10
Prevent endless loop in lock cycle detection #1510

trancexpress:
This change prevents an endless loop in: ReentrantCycleDetectingLock.detectPotentialLocksCycle()

Due to how code in ReentrantCycleDetectingLock.lockOrDetectPotentialLocksCycle() is synchronized,
its possible for a thread to both own/hold a lock (according to ReentrantCycleDetectingLock.lockOwnerThread)
and wait on the same lock (according to CycleDetectingLock.lockThreadIsWaitingOn).
In this state, if another thread tries to hold the same lock an endless loop will occur when calling detectPotentialLocksCycle().

With this change detectPotentialLocksCycle() removes the lock owning thread from
ReentrantCycleDetectingLock.lockOwnerThread,
if it detects that "this" lock is both waited on and owned by the same thread.
This prevents the endless loop during cycle detection.

Fix for: #1510

### DIFF
--- a/core/src/com/google/inject/internal/CycleDetectingLock.java
+++ b/core/src/com/google/inject/internal/CycleDetectingLock.java
@@ -259,6 +259,11 @@ interface CycleDetectingLock<ID> {
             // owner thread depends on current thread, cycle detected
             return potentialLocksCycle;
           }
+          // Its possible the lock owner thread has not removed itself yet from the waiting-on-lock list.
+          // So we remove it here, to prevent an endless loop. See #1510.
+          if (lockOwnerWaitingOn == this) {
+            lockThreadIsWaitingOn.remove(this.lockOwnerThread);
+          }
         }
         // no dependency path from an owner thread to a current thread
         return ImmutableListMultimap.of();


trancexpress:
OK I understand the problem now. ```ReentrantCycleDetectingLock.lockOrDetectPotentialLocksCycle()``` has the following code:

```
        final Thread currentThread = Thread.currentThread();
        synchronized (CycleDetectingLockFactory.class) {
          checkState();
          // Add this lock to the waiting map to ensure it is included in any reported lock cycle.
          lockThreadIsWaitingOn.put(currentThread, this);
          ListMultimap<Thread, ID> locksInCycle = detectPotentialLocksCycle();
          if (!locksInCycle.isEmpty()) {
            // We aren't actually going to wait for this lock, so remove it from the map.
            lockThreadIsWaitingOn.remove(currentThread);
            // potential deadlock is found, we don't try to take this lock
            return locksInCycle;
          }
        }

        // this may be blocking, but we don't expect it to cause a deadlock
        lockImplementation.lock();

        synchronized (CycleDetectingLockFactory.class) {
          // current thread is no longer waiting on this lock
          lockThreadIsWaitingOn.remove(currentThread);
          checkState();

          // mark it as owned by us
          lockOwnerThread = currentThread;
          lockReentranceCount++;
          // add this lock to the list of locks owned by a current thread
          locksOwnedByThread.put(currentThread, this);
        }
```

What this code (I assume) wants to do is:

1. add the current thread and "this" lock to ```lockThreadIsWaitingOn```
2. check for cycles
2.1. this results in calling ```detectPotentialLocksCycle()``` and then calling ```addAllLockIdsAfter()```
2.2. the two methods will loop forever, if they iterate over a thread that both owns a lock and waits on that lock, see the respective code:
```
        /* in detectPotentialLocksCycle() */
        ReentrantCycleDetectingLock<?> lockOwnerWaitingOn = this;
        // try to find a dependency path between lock's owner thread and a current thread
        while (lockOwnerWaitingOn != null && lockOwnerWaitingOn.lockOwnerThread != null) {
          Thread threadOwnerThreadWaits = lockOwnerWaitingOn.lockOwnerThread;
          // in case locks cycle exists lock we're waiting for is part of it
          lockOwnerWaitingOn =
              addAllLockIdsAfter(threadOwnerThreadWaits, lockOwnerWaitingOn, potentialLocksCycle);
```
```
        /* in addAllLockIdsAfter() */
        ReentrantCycleDetectingLock<?> unownedLock = lockThreadIsWaitingOn.get(thread);
        // If this thread is waiting for a lock add it to the cycle and return it
        if (unownedLock != null && unownedLock.lockFactory == this.lockFactory) {
          @SuppressWarnings("unchecked")
          ID typed = (ID) unownedLock.userLockId;
          potentialLocksCycle.put(thread, typed);
        }
        return unownedLock;
```
4. remove the current thread and "this" from ```lockThreadIsWaitingOn```

If the entire sequence was done in 1 synchronized block, there wouldn't be a problem. However, ```CycleDetectingLockFactory.class``` is locked to first add to ```lockThreadIsWaitingOn``` and check for cycles. Then the synchronized block is left, and another synchronized block is entered, to remove from ```lockThreadIsWaitingOn```.

What instead can occur (and occurs in the reproducer), is that 1 thread adds to ```lockThreadIsWaitingOn``` with a lock and then acquires the lock (and is so owning the lock). Then another thread does a cycle check, wanting to acquire the same lock. It sees the owner thread of the lock, tries to check for cycles with it and loops forever. While the first thread waits on removing from ```lockThreadIsWaitingOn``` - which would prevent the endless loop.

Unfortunately I don't know whether the adding the removal from ```lockThreadIsWaitingOn``` in the first synchronized block breaks other code. It might be safer to detect the bad case scenario and just bail out (as in the suggested workaround). I don't know what this means for the cycle detection though. The test added for #915 passes, but seeing that it misses the endless loop reported in #1510 maybe its not extensive enough to catch problems that this change introduces.

core/src/com/google/inject/internal/CycleDetectingLock.java
```diff
@@ -249,6 +249,9 @@ private ListMultimap<Thread, ID> detectPotentialLocksCycle() {
             MultimapBuilder.linkedHashKeys().arrayListValues().build();
         // lock that is a part of a potential locks cycle, starts with current lock
         ReentrantCycleDetectingLock<?> lockOwnerWaitingOn = this;
+        // Its possible the lock owner thread has not removed itself yet from the waiting-on-lock list.
+        // So we remove it here, to prevent an endless loop. See #1510.
+        lockThreadIsWaitingOn.remove(lockOwnerWaitingOn.lockOwnerThread);
```
iloveeclipse, line 254: `this.remove(this.lockOwnerThread);`  ?

trancexpress: You mean ```lockThreadIsWaitingOn.remove(this.lockOwnerThread)```?

Sure, why not. I prefer to use ```lockOwnerWaitingOn``` after this is stored into it, but its whichever.

trancexpress:
From my POV this is ready for a merge, IMO it fixes the concurrency problem in a reliable and simple way.

trancexpress:
Looks like there are fails in the cycle detection test, will have to check why:

```
Tests in error: 
  testCycleDetectingLockFactoriesDoNotDeadlock(com.google.inject.internal.CycleDetectingLockTest)
  testCycleReporting(com.google.inject.internal.CycleDetectingLockTest)
```

iloveeclipse:
@andrewthehan, @sameb, @cpovirk, @markmarch, **anyone** : could we please move forward with review?

PR #1639, opened 2022-06-30
fix: corrected two grammar issues

Jan-Brunner:
I added two missing hyphen's


sameb:
Thank you for this PR, @Jan-Brunner. I don't think we need to add the hyphens, so am closing this without merging.

Issue #700, opened 2014-07-07
binding order matters when requireExplicitBindings is on


gissuebot:
_From [pmoor@google.com](https://code.google.com/u/111568174877005671726/) on April 11, 2012 19:18:52_

See the attached test case, it fails with

1) Explicit bindings are required and com.google.apps.framework.producers.GuicePocTest$A is not explicitly bound.
&nbsp;&nbsp;while locating com.google.apps.framework.producers.GuicePocTest$A
&nbsp;&nbsp;&nbsp;&nbsp;for parameter 0 at com.google.apps.framework.producers.GuicePocTest$BImpl.&lt;init&gt;(GuicePocTest.java:17)
&nbsp;&nbsp;at com.google.apps.framework.producers.GuicePocTest$1.configure(GuicePocTest.java:25)

When reversing the binding order (first A, then B), it passes.

**Attachment:** [gist](https://gist.github.com/5d62c7df985395ca3cd4)
&nbsp;&nbsp;&nbsp;_[GuicePocTest.java](https://gist.githubusercontent.com/gissuebot/5d62c7df985395ca3cd4/raw/c10868923ac1aa5d1245065fb487eabb978a8bbd/GuicePocTest.java)_

_Original issue: http://code.google.com/p/google-guice/issues/detail?id=700_

PR #1650, opened 2022-10-08
Fix of #700

nineninesevenfour:
Ensure the order of calls to Binder.bind(...) in combination with @ does not affect whether classes are bound explicitly or with jIT binding.

Signed-off-by: Harald Fassler <harald.fassler+9974@gmail.com>

### DIFF
--- a/core/src/com/google/inject/internal/InjectorImpl.java
+++ b/core/src/com/google/inject/internal/InjectorImpl.java
@@ -56,7 +56,7 @@ import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.util.Collections;
-import java.util.HashSet;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


nineninesevenfour:
Hello, I am a new contributor 👋, let me explain, what I did an why.

The issue popped up through https://stackoverflow.com/questions/73901723.

After some tests I found out that the issue did surprisingly not occur in combination with `@ProvidedBy`. After some debugging I came to the conclusion that the solution would involve delayed initialization with `DelayedInitialize`.  I hesitated to create a distinct `LinkedImplentedByBindingImpl` and instead extended `LinkedBindingImpl` letting it implement `DelayedInitialize`. Also I did not change the way the binding is initialized and kept the call to `ProxyFactory.notify(...)`, but changed it to a lambda expression. I ignored the fact, that `DelayedInitialize` might receive a different `Injector` instance. But through temporary added code (removed before committing again), I found no test that would come up with a different `Injector` instance here. Maybe you could guide me to find a case, where this could happen.

Unfortunately the change made `InjectorImpl.cleanup()` fail in context of `ImplicitBindingTest.testCircularJitBindingsLeaveNoResidue()`. This was because the `cleanup` method relied on the order of encountered dependencies of injectees. Therefor I changed `encountered` from a set to a map,  additionally  remembering if a previous encounter failed or not.

In total I tried to make the change as unobtrusive as possible.

sameb:
Thank you for contributing this, @nineninesevenfour !  The change is in a somewhat tricky area of the code that I wrote many many years ago, so I will review it more closely next week and potentially tweak it a bit before submitting the fix.

PR #1665, opened 2023-02-07
fix: prevent NullPointerException if targetPackageName is null

PLin2023:


Stefan-Hintz (approved):

sameb:
Thanks for submitting this @PLin2023 . However, the proposed change does not actually prevent any NPEs. It just changes the NPE to be more informative: telling you that "targetPackageName" is what caused the NPE. As of Java 17, NPEs automatically tell you what was null.  See https://openjdk.org/jeps/358 (or other articles explaining it, like https://dev.to/harsvnc/the-new-nullpointer-exception-in-java-17-4bcb).

If you'd like to resubmit a new PR that actually _ignores_ null packages, please explain in what scenario you encounter a null package that you'd like to ignore, and include a test.  Thanks!

PR #1708, opened 2023-04-24
Completely remove OSGi Import-Package 'javax.annotation'

HannesWell:
As suggested (and I assume also intended) in https://github.com/google/guice/pull/1173 this PR changes the configuration of the maven-bundle-plugin to completely remove the import of the package `javax.annotation`, which seems not to be required at runtime.

With PR https://github.com/google/guice/pull/1173 respectively https://github.com/google/guice/pull/1697 the OSGi package import of `javax.annotation;version="[3.0,4)"` was only changed to `javax.annotation`, i.e. the version range was removed.

@mcculls do you want to have a look at this?

Additionally update to latest maven-bundle-plugin 5.1.8.

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -149,7 +149,7 @@
         <configuration>
           <instructions>
             <Bundle-Name>${project.artifactId}</Bundle-Name>
-            <Import-Package>!org.objectweb.asm.*,!com.google.inject.*,*</Import-Package>
+            <Import-Package>!org.objectweb.asm.*,!com.google.inject.*,!javax.annotation,*</Import-Package>
             <Eclipse-ExtensibleAPI>true</Eclipse-ExtensibleAPI>
           </instructions>
         </configuration>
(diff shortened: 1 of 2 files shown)


sameb (approved):

HannesWell:
Thank you @sameb for the quick review and merge.
Can you tell when the next release of guice that contains this change will approximately be done?

sameb:
Sure thing!  Thanks for the PR!

My hope is to finish merging outstanding pull requests this week and do a cursory sweep over open issues, fixing anything reasonable.

The biggest barrier to a release is a resolution to #1383.  At this point, I think we mostly need a Maven expert who can advise on how to structure the POMs to allow pushing an artifact with different deps and bytecode, so we can do a final 6.0 release as-is, and then a 7.0 release using the jakarta namespace as deps (despite the code still referencing javax).  As soon as that's settled, I can cut a new release.

HannesWell:
Thanks for the elaboration.

Looks like you already found a solution/plan. Looking forward to the new release!

PR #1714, opened 2023-04-28
Migrate tests to JUnit 5

kashike:
This pull request migrates tests over to JUnit 5, rather than a mix of JUnit 3 and JUnit 4 tests - the only exception being the Tck tests.

All tests pass, with the exception of a couple Tck tests which fail on master as well, when running with Maven. I've been unable to figure out how to get tests to run on Bazel yet.

<sub>I'm unsure if this is something that the Guice team would merge (anytime soon, or at all), but I had a bit of spare time on my hands. 😄</sub>


sameb:
Thank you very much for this PR.  Unfortunately, Google's internal repo doesn't support JUnit5, so we can't merge it in.  I recommend following https://github.com/bazelbuild/bazel/issues/6681 -- if/when Bazel supports JUnit5 natively, that's likely a good sign that we've added support internally for junit5. Once we're able to accept this, I'd be glad to accept a new PR with these changes.

PR #1716, opened 2023-04-29
Implement `and` and `or` in `Matcher` directly

kashike:
#1177 

This makes `AbstractMatcher` itself empty now - do you want it to be marked as deprecated?

### DIFF
--- a/core/src/com/google/inject/matcher/AbstractMatcher.java
+++ b/core/src/com/google/inject/matcher/AbstractMatcher.java
@@ -16,88 +16,24 @@
 
 package com.google.inject.matcher;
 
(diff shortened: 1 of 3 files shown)


sameb:
I'm a little concerned that removing the overrides from AbstractMatcher will break binary compatibility with old jars that delegate to AbstractMatcher.{and,or}. WDYT of keeping those overrides (even though they're redundant on the source-code level)?

Marking as @Deprecated SGTM, with the javadoc explaining how implementing Matcher is enough (because of the default `and` and `or` methods).

sameb (approved):

Issue #1121, opened 2017-07-25
Lacking injection point information when using the deprecated DependenyChain replacement.


Cybermaxke:
The provided example to replace the deprecated dependency chain method doesn't provide enough context about which field or parameter is being injected. I tried to replace dependency chain method, but the most accurate I got was knowing which method/constructor/class is being injected.

Knowing which field/parameter is being injected allowed `Provider`s to return values based on the annotated elements, without having to bind each possible combination.

The dependency chain solution can be found in the following PR changes: https://github.com/SpongePowered/SpongeCommon/pull/1435/files/1bd19db6f0b2363b8bc72960df0b0947e04cc1e0
A custom `InjectionPoint` is constructed from the information in the dependency chain, this object only exposes the annotations/type/target type of the field/parameter. There is also a small example available in the PR.

Then I was looking for a different solution for the dependency chain and ended up stuck in the following changes:
https://github.com/LanternPowered/LanternServer/commit/f67d3e86af549b7e64c43f5f0de85d6fe5aaf211

Not sure what the best solution could be to resolve the issue, undeprecating the `getDependencyChain` method is an option. Or similar to the `InjectionPoint` injection, could it also be done for the `Dependency` by guice itself, making it unnecessary to scan the chain. Maybe you guys can find a better solution for this issue. Maybe there is even a different solution that I haven't been able to find.

kashike:
I'm going to give this a bump - any solution known? If not, is it possible to have this reverted?

kashike:
I hate to be one of *those* people who bumps so often, but it'd be great to get a response on this - @dimo414?

dimo414:
@kashike sorry, I'm afraid I don't have any insights here.

Cybermaxke:
The DependencyChain got deprecated in the following commit: https://github.com/google/guice/commit/99c1047b23af62d17b41bda03947a54f581b21c8
And one of the reasons it being depecrated was not being "popular" enough and guice being more performant when removing it (`A benchmarking experiment* shows that maintaining the dependency stack costs about 10% of guice performance.  Anecdotally this feature also isn't very popular, so eliminating it seems reasonable.`). Maybe a opt-in solution would be better instead of removing it?

Or maybe a method to retrieve the dependency chain without the source (would reduce the internal dependency stack size by half: https://github.com/google/guice/blob/master/core/src/com/google/inject/internal/InternalContext.java#L46).

kashike:
@lukesandberg - input?

[...]

PR #1721, opened 2023-05-03
Introduce a new type of provider with access to the injection point

kashike:
This is an attempt to re-introduce functionality that was removed due to the changes made in https://github.com/google/guice/commit/6f185140d88feafeec65039a213f08a0abeff268.

ref #1121

If anyone has feedback about this, please feel free to leave a comment.

### DIFF
--- a/core/src/com/google/inject/internal/ContextualProviderInstanceBindingImpl.java
+++ b/core/src/com/google/inject/internal/ContextualProviderInstanceBindingImpl.java
@@ -103,7 +103,7 @@ class ContextualProviderInstanceBindingImpl<T> extends BindingImpl<T> implements
     @Override
     public T get(InternalContext context, Dependency<?> dependency, boolean linked) {
       InjectionContext injectionContext = InjectionContext.fromDependency(dependency);
-      return context == null
+      return injectionContext == null
           ? get() // without context
           : provider.get(injectionContext);
     }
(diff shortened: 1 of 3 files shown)


sameb:
HI @kashike -- FYI, I'm focusing my efforts right now on getting a 6.0 & 7.0 release out the door, and this PR won't be in those releases.

IIUC, you'd like to change what a Provider creates based on arbitrary data associated with the injection point that's being injected.  That has thus far been an explicit anti-goal of Guice. If the returned type can change based on who's requesting it, that makes reasoning about the object graph much harder.

Additionally, as far as quirks of the API go, I can't imagine how this would work with scopes. As-is, it looks like we'll capture whatever the first return value is & replay it to other injection points. That definitely feels wrong, though I don't know what "right" should even be.

sameb:
> That is what the goal of this pull request is, for the most part. This functionality used to be possible in Guice prior to the release of 4.4 when [6f18514](https://github.com/google/guice/commit/6f185140d88feafeec65039a213f08a0abeff268) was pushed (although, in a slightly hacky way), and the Guice team has been radio silent on #1121 since 2017.

The ability to do it before was my mistake. When adding the ProvisionListener/ProvisionInvocation API, it didn't occur to me that folks would have their Provider impls refer to data stashed from the ProvisionListener and change the way the Provider works.

I agree this PR is a relatively clean way to do what you're proposing... but I don't think we can incorporate it into Guice for the reasons explained in the prior comment: it goes against the philosophy, and it doesn't fit cleanly into the model (e.g, the weird interaction with scopes).

Issue #1767, opened 2023-10-14
8 Flaky tests in the module extensions/throwingproviders


harshith2000:
Found 8 flaky tests in the module extensions/throwingproviders

1. Tests that fail

Module: extensions/throwingproviders

Class Name1: com.google.inject.throwingproviders.CheckedProvidersTest, 
Test Name: [testUnsupportedMethods_otherMethod_throwsIllegalArgumentException](https://github.com/google/guice/blob/73dc9f3b8afac8b742b766049afc69caf95dbf51/extensions/throwingproviders/test/com/google/inject/throwingproviders/CheckedProvidersTest.java#L78)

Class Name2: testUnsupportedMethods_otherMethod_throwsIllegalArgumentException
[...]

PR #1768, opened 2023-10-17
Fixed 8 Flaky tests in the module extensions/throwingproviders

harshith2000:
This PR fixes the issue [#1767](https://github.com/google/guice/issues/1767)

Fix:
Since the order of the methods or error messages is not guaranteed, it is important to consider all possible orders. Fortunately, only two possible orders are possible for both methods and error messages. By checking if the actual message is equal to any of the orders, we can fix the flakiness.


sameb:
these tests aren't flaky in practice, and there's no need to change the tests based on theoretical behaviors.

harshith2000:
> these tests aren't flaky in practice, and there's no need to change the tests based on theoretical behaviors.

Thank for the feedback.

PR #1777, opened 2023-11-28
design and implementation smells removed

HarshMehta19:
design and implementation smells removed


sameb:
there's no compelling reason for this change AFAICT.  closing PR.  please provide more explanation of why you want to change this if you re-open.

PR #1795, opened 2024-02-05
Bump dependency ASM to 9.6

gruenich:
It was ASM 9.5


hashhar:
might want to also bump the value at https://github.com/google/guice/blob/fd77c955a9a133df87bf0b7c5b359819617e0a05/core/src/com/google/inject/internal/util/LineNumbers.java#L50, or not?

ryantomlinson95:
Related issue: https://github.com/google/guice/issues/1822
This would be great to have since it causes issues with both Java 22 and 23 at the moment.

PR #1840, opened 2024-11-01
Add source details to InternalProvisionException in ExposedKeyFactory

panic08:
Here I am improving the error traceability in ExposedKeyFactory by adding source details to the InternalProvisionException when it occurs. The exception message now includes specific keys and private elements, making it easier to determine the source of the error


lukesandberg:
I believe i resolved in this https://github.com/google/guice/commit/b673643f6a06e66d2dffbe0f0a703c91e89e2b94

PR #1848, opened 2024-12-07
Change the isPresent Option state to functional style

panic08:
Here we switch to the one expression functional style Stream api, which looks more readable


panic08:
Closed due to a hidden character in a branch and added in another PR #1849

PR #1892, opened 2025-03-27
Update ASM to 9.7.1

wendigo:
This version supports Java up to version 25.


wendigo:
cc @lukesandberg

thmarx:
+1 for this PR

wendigo:
@thmarx protip: you can use non-shaded guice using `classes` classifier (which will pull ASM version that you want)

thmarx:
@wendigo yes, that's the workaround we are using at the moment.

PR #1894, opened 2025-04-02
Finish the experimental method handle support

lukesandberg:
This series addresses known issues in the experimental method handle implementation

  - MethodHandles produced by InternalFactory.getHandle are now cached.  This should speed up linking and reduce the number of 'lambda forms' we generate.
      - Doing this required demoting `Dependency` from the `getHandle` to be a runtime method parameter and changing InternalFactory to be an abstract class
  - We no longer generate custom `Provider` subclasses to hold `invokedynamic` instructions and instead use a simple hand written Provider class that lazily allocates a `MethodHandle` and calls invokeExact.
       - This should greatly speed up injector creation at the cost of a slightly more expensive `invokeExact` call. 
   - requestStaticInjection and MembersInjectors are now fully supported with MethodHandles.
   
 I imagine there are some as yet undiscovered issues but this should unblock broader experimentation.

### DIFF
--- a/core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
+++ b/core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
@@ -136,6 +136,8 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
               .asType(methodType(Object.class, Object.class, Object[].class));
       // (Object[])->Object
       handle = MethodHandles.insertArguments(handle, 0, (Object) null); // no receiver type.
+      // catch here so we don't catch errors from our parameters
+      handle = InternalMethodHandles.catchErrorInConstructorAndRethrowWithSource(handle, injectionPoint);
       // (InternalContext)->Object
       handle =
           MethodHandles.filterArguments(
@@ -165,6 +167,9 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
       // See comments in ProviderMethod on how this rarely happens and why it happens
       // (Object[])->Object
       var handle = CONSTRUCTOR_NEWINSTANCE_HANDLE.bindTo(constructor);
+      // Catch here so we don't catch errors from our parameters
+      handle = InternalMethodHandles.catchErrorInConstructorAndRethrowWithSource(handle, injectionPoint);
+
       // (InternalContext)->Object
       handle =
           MethodHandles.filterArguments(
@@ -201,7 +206,9 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
           IntStream.range(0, parameterHandles.length)
               .mapToObj(i -> castReturnTo(parameterHandles[i], type.parameterType(i)))
               .toArray(MethodHandle[]::new);
-      var handle = MethodHandles.filterArguments(target, 0, typedHandles);
+      // catch errors from the constructor
+      var handle =  InternalMethodHandles.catchErrorInConstructorAndRethrowWithSource(target, injectionPoint);
+      handle = MethodHandles.filterArguments(handle, 0, typedHandles);
       handle = castReturnToObject(handle); // satisfy the signature of the factory type.
       return MethodHandles.permuteArguments(
           handle, InternalMethodHandles.ELEMENT_FACTORY_TYPE, new int[typedHandles.length]);
--- a/core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
+++ b/core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
@@ -144,7 +144,7 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
       // merge all the internalcontext parameters into a single object factory.
       handle =
           MethodHandles.permuteArguments(
-              handle, InternalMethodHandles.FACTORY_TYPE, new int[parameterHandles.length]);
+              handle, InternalMethodHandles.ELEMENT_FACTORY_TYPE, new int[parameterHandles.length]);
       return handle;
     }
   }
@@ -175,7 +175,7 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
       handle = MethodHandles.filterArguments(handle, 0, parameterHandles);
       // merge all the internalcontext parameters into a single object factory.
       return MethodHandles.permuteArguments(
-          handle, InternalMethodHandles.FACTORY_TYPE, new int[parameterHandles.length]);
+          handle, InternalMethodHandles.ELEMENT_FACTORY_TYPE, new int[parameterHandles.length]);
     }
   }
 
@@ -210,7 +210,7 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
       var handle = MethodHandles.filterArguments(target, 0, typedHandles);
       handle = castReturnToObject(handle); // satisfy the signature of the factory type.
       return MethodHandles.permuteArguments(
-          handle, InternalMethodHandles.FACTORY_TYPE, new int[typedHandles.length]);
+          handle, InternalMethodHandles.ELEMENT_FACTORY_TYPE, new int[typedHandles.length]);
     }
   }
 }
--- a/core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
+++ b/core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
@@ -16,7 +16,6 @@
 
 package com.google.inject.internal;
 
-import static com.google.inject.internal.InternalMethodHandles.CONSTRUCTOR_NEWINSTANCE_HANDLE;
 import static com.google.inject.internal.InternalMethodHandles.castReturnTo;
 import static com.google.inject.internal.InternalMethodHandles.castReturnToObject;
 import static java.lang.invoke.MethodType.methodType;
@@ -166,7 +165,7 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
     public MethodHandle getConstructHandle(MethodHandle[] parameterHandles) {
       // See comments in ProviderMethod on how this rarely happens and why it happens
       // (Object[])->Object
-      var handle = CONSTRUCTOR_NEWINSTANCE_HANDLE.bindTo(constructor);
+      var handle = InternalMethodHandles.newInstanceHandle(constructor);
       // Catch here so we don't catch errors from our parameters
       handle = InternalMethodHandles.catchErrorInConstructorAndRethrowWithSource(handle, injectionPoint);
 
--- a/core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
+++ b/core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
@@ -134,17 +134,12 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
           InternalMethodHandles.BIFUNCTION_APPLY_HANDLE
               .bindTo(fastConstructor)
               .asType(methodType(Object.class, Object.class, Object[].class));
+      // (Object[])->Object
       handle = MethodHandles.insertArguments(handle, 0, (Object) null); // no receiver type.
-      // NOTE: is is safe to use asCollector here because the number of parameters is the same
-      // as the number of parameters to the constructor which should never exceed the maxiumum
-      // number of method parameters.
-      handle = handle.asCollector(Object[].class, parameterHandles.length);
-      // Pass all the parameters to the constructor.
-      handle = MethodHandles.filterArguments(handle, 0, parameterHandles);
-      // merge all the internalcontext parameters into a single object factory.
+      // (InternalContext)->Object
       handle =
-          MethodHandles.permuteArguments(
-              handle, InternalMethodHandles.ELEMENT_FACTORY_TYPE, new int[parameterHandles.length]);
+          MethodHandles.filterArguments(
+              handle, 0, InternalMethodHandles.buildObjectArrayFactory(parameterHandles));
       return handle;
     }
   }
@@ -168,14 +163,13 @@ final class DefaultConstructionProxyFactory<T> implements ConstructionProxyFacto
     @Override
     public MethodHandle getConstructHandle(MethodHandle[] parameterHandles) {
       // See comments in ProviderMethod on how this rarely happens and why it happens
+      // (Object[])->Object
       var handle = CONSTRUCTOR_NEWINSTANCE_HANDLE.bindTo(constructor);
-      // collect the parameters into an array of type Object[]
-      handle = handle.asCollector(Object[].class, parameterHandles.length);
-      // apply all the parameters to the constructor.
-      handle = MethodHandles.filterArguments(handle, 0, parameterHandles);
-      // merge all the internalcontext parameters into a single object factory.
-      return MethodHandles.permuteArguments(
-          handle, InternalMethodHandles.ELEMENT_FACTORY_TYPE, new int[parameterHandles.length]);
+      // (InternalContext)->Object
+      handle =
+          MethodHandles.filterArguments(
+              handle, 0, InternalMethodHandles.buildObjectArrayFactory(parameterHandles));
+      return handle;
     }
   }
 
--- a/core/src/com/google/inject/internal/InjectionRequestProcessor.java
+++ b/core/src/com/google/inject/internal/InjectionRequestProcessor.java
@@ -138,8 +138,6 @@ final class InjectionRequestProcessor extends AbstractProcessor {
     }
 
     void injectMembers() {
-
-      // TODO(b/366058184): investigate a methodhandle version of this.
       InternalContext context = injector.enterContext();
       try {
         boolean isStageTool = injector.options.stage == Stage.TOOL;
@@ -147,10 +145,27 @@ final class InjectionRequestProcessor extends AbstractProcessor {
           // Run injections if we're not in tool stage (ie, PRODUCTION or DEV),
           // or if we are in tool stage and the injection point is toolable.
           if (!isStageTool || memberInjector.getInjectionPoint().isToolable()) {
-            try {
-              memberInjector.inject(context, null);
-            } catch (InternalProvisionException e) {
-              errors.merge(e);
+            if (InternalFlags.getUseExperimentalMethodHandlesOption()) {
+              try {
+                // In theory, constructing the handle to invoke it exactly once is expensive and wasteful, and it is true for
+                // directly injecting the member this is probably slower than the reflective SingleFieldInjector. However,
+                // by taking this path we::
+                // 1. Don't need to construct fastclasses for method injections (SingleMethodInjector).
+                // 2. construct fast classes for transitive injections (constructors/@Provides methods).
+                // 3. Can leverage or initialize caches for transitive InternalFactory.getHandle calls.
+                memberInjector.getInjectHandle(new LinkageContext()).invokeExact((Object) null, context);
+              } catch (InternalProvisionException e) {
+                errors.merge(e);
+              } catch (Throwable t) {
+                // This will propagate unexpected Errors.
+                InternalMethodHandles.sneakyThrow(t);
+              }
+            } else {
+              try {
+                memberInjector.inject(context, null);
+              } catch (InternalProvisionException e) {
+                errors.merge(e);
+              }
             }
           }
         }
--- a/core/src/com/google/inject/internal/InternalFactory.java
+++ b/core/src/com/google/inject/internal/InternalFactory.java
@@ -21,7 +21,6 @@ import static com.google.common.base.Preconditions.checkNotNull;
 import com.google.inject.Provider;
 import com.google.inject.spi.Dependency;
 import java.lang.invoke.MethodHandle;
-import java.lang.invoke.MethodHandles;
 
 /**
  * Creates objects which will be injected.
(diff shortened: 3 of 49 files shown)


sameb:
Thanks Luke!  Will try to pull and test, and will review here soon!

lukesandberg:
Welp I couldn't get bazel working and it shows... lol

lukesandberg:
Fixed the `@ForOverride` issue that the workflow flagged

sameb (approved):
(approving so i can pull and validate internally, still need to review properly)

core/src/com/google/inject/internal/DefaultConstructionProxyFactory.java
```diff
@@ -134,17 +134,12 @@ public MethodHandle getConstructHandle(MethodHandle[] parameterHandles) {
...
+      // (Object[])->Object
       handle = MethodHandles.insertArguments(handle, 0, (Object) null); // no receiver type.
-      // NOTE: is is safe to use asCollector here because the number of parameters is the same
-      // as the number of parameters to the constructor which should never exceed the maxiumum
-      // number of method parameters.
-      handle = handle.asCollector(Object[].class, parameterHandles.length);
-      // Pass all the parameters to the constructor.
-      handle = MethodHandles.filterArguments(handle, 0, parameterHandles);
-      // merge all the internalcontext parameters into a single object factory.
+      // (InternalCotnext)->Object
```
sameb, line 139: s/Cotnext/Context

core/src/com/google/inject/internal/InjectionRequestProcessor.java
```diff
@@ -138,19 +138,34 @@ void validate() {
...
           // Run injections if we're not in tool stage (ie, PRODUCTION or DEV),
           // or if we are in tool stage and the injection point is toolable.
           if (!isStageTool || memberInjector.getInjectionPoint().isToolable()) {
-            try {
-              memberInjector.inject(context, null);
-            } catch (InternalProvisionException e) {
-              errors.merge(e);
+            if (InternalFlags.getUseExperimentalMethodHandlesOption()) {
+              try {
+                // In theory, constructing the handle to invoke it exactly once is expensive and wasteful, and it is true for
```
sameb, line 150: Question: Why does this need to be a one-off instead of being reusable like the other paths?

core/src/com/google/inject/internal/InternalFactory.java
```diff
@@ -21,14 +21,21 @@
...
-interface InternalFactory<T> {
+abstract class InternalFactory<T> {
+
+  /**
+   * This will be a {@link MethodHandle} if the handle is always cachable, or a {@code
+   * MethodHandle[2]} if the handle is conditionally cachable based on the linked status. In which
+   * case index 0 is the handle for when not linked and index 1 is for when it is linked.
+   * 
+   */
+  private volatile Object handleCacheResult;
```
sameb, line 38: Would it be wildly inefficient to make this more type-safe, some kind of MethodHandleRef interface type w/ implementations: SimpleRef and LinkedRef, where SimpleRef is just a passthru & LinkedRef has two volatile fields of linkedRef & unlinkedRef?

(This would also negate the question below about cross-thread safety of setting items in the array.)

core/src/com/google/inject/internal/InternalFactory.java
```diff
@@ -53,14 +60,127 @@ default Provider<T> makeProvider(InjectorImpl injector, Dependency<?> dependency
...
+            var local = handleCacheResult;
+            if (local != null) {
+              if (local instanceof MethodHandle[]) {
+                var array = (MethodHandle[]) local;
+                var handle = array[index];
+                if (handle != null) {
+                  // this implies we lost a race
+                  return handle;
+                }
+                array[index] = result.methodHandle;
```
sameb, line 169: Is setting an item in an array a safe cross-thread operation (even if the array is volatile)?  I thought volatility was for the field itself, not the contents within it.  (Same below @ L177)

core/src/com/google/inject/internal/InternalMethodHandles.java
```diff
@@ -865,436 +879,118 @@ public String toString() {
...
-            );
-        mv.visitEnd();
-      }
-      // We must define the class as collectable so that it can be garbage collected.  The static
-      // field will hold a reference to an InternalFactory which may hold references to the
-      // injector.
-      Class<?> clazz;
-      try {
-        clazz =
-            ClassDefining.defineCollectable(
```
sameb, line 1000: Should we also delete all the ClassDefining stuff introduced in the earlier changes ?  AFAICT we don't need "collectable" infra anymore.

core/src/com/google/inject/internal/LinkageContext.java
```diff
@@ -58,30 +60,31 @@ final class LinkageContext {
...
       previous = new MutableCallSite(InternalMethodHandles.FACTORY_TYPE);
-      linkingFactories.put(source, previous);
+      linkingFactories.put(factory, previous);
     }
     if (previous instanceof MutableCallSite) {
       // We are re-entering the same factory, so we can just return the dynamic invoker and rely on
       // the original invocation to finish the construction and call setTarget to finalize the
       // callsite.
-      return ((MutableCallSite) previous).dynamicInvoker();
+      return InternalFactory.makeUncachable(((MutableCallSite) previous).dynamicInvoker());
```
sameb, line 75: Can you elaborate in comments on why this is uncachable?

core/src/com/google/inject/spi/BindingSourceRestriction.java
```diff
@@ -345,8 +345,12 @@ private static Stream<Class<? extends Annotation>> getPermits(Class<?> clazz) {
           Stream.concat(
               annotations, Arrays.stream(clazz.getAnnotatedSuperclass().getAnnotations()));
     }
-    return annotations
```
sameb, line 348: Why change?

sameb (commented):
General question: I'm curious about the caching strategy. AIUI, the cache for a given MH is unique per InternalFactory, but InternalFactory isn't widely shared (they are unique per binding IIRC). Is this improving cachability (and repeated injector constructions in tests, etc..) because we're removing the custom bytecode-generated Provider class?  It seems like each distinct Injector will still have its own unique MethodHandles... but I guess that's more lightweight than a whole class definition per Provider?  What value is the MH cacheability actually giving us if its unique per binding (or am I misunderstanding the scope of the caching)?

sameb:
(Also, I ran through the depot last night, but forgot to flip the "enable this" flag to true... so running again now. Will see what it shakes out.)

core/src/com/google/inject/internal/InjectionRequestProcessor.java
lukesandberg: AFAICT this path is only hit once since this is for static injection, so there wouldn't be much of a benefit of combining all the handles together.

sameb:
> (Also, I ran through the depot last night, but forgot to flip the "enable this" flag to true... so running again now. Will see what it shakes out.)

depot is fairly happy.  Some things that surfaced:

* Behavior changed for Provider instances that throw Errors (and possibly other things that throw errors?).  e.g, something like `bind(Foo.class).toProvider( { throw new AssertionError() } )`.. previously `injector.getInstance(Foo.class)` would surface that `AssertionError`, whereas now its wrapped inside a `ProvisionException`.  I'd argue that the `ProvisionException` is actually _more correct_, but we should avoid the behavior change for now and do it intentionally later if desired.  (This is commonly surfaced from @Provides methods that threw Errors, but sometimes also just raw Provider instances. Presumably it'd be the same with a constructor that threw this way w/o any custom provider instance?)

* A very specific allocation leakage test which _should_ have shaken out "warmup" allocations, but now there appears to be new long-lived allocations per usage (despite a long-lived injector). I don't *think* this PR is the cause (because AFAICT the "warmup" phase should pre-create all the requisite MHs), but will need to dig in further.  Not blocking this PR, but maybe you see something in this PR that's always creating new data per injection.

core/src/com/google/inject/internal/InternalMethodHandles.java
lukesandberg: Yes. Will delete!

core/src/com/google/inject/spi/BindingSourceRestriction.java
lukesandberg: I kept getting a build error here from maven

```
Type mismatch: cannot convert from Stream<Class<capture#26-of ? extends Annotation>> to Stream<Class<? extends Annotation>>Java(16777235)
```

Introducing this expliocit cast resolved it, so i guess i cannot figure out why this isn't a problem at head?

im guessing it is because i am using a different version of `javac` 

```
% javac --version
javac 21.0.6
```

core/src/com/google/inject/internal/LinkageContext.java
lukesandberg: Added a comment.  There is no correct caching setting here so we don't try, when the original invocation returns we will cache that result.

core/src/com/google/inject/internal/InternalFactory.java
lukesandberg: Done.  I was concerned about adding memory overhead to the case where methodhandles were not in in use but i believe i found a solution to that.  This should resolve your (legitimate) threadsafety concerns as well.

Basically we can create a small type hierarchy that can handle the different transitions and then a simple compare-and-swap to pick a winner.  The cost is allocating some tiny objects but since we onlyt do it when caching, i don't really think it will matter.

lukesandberg:
> General question: I'm curious about the caching strategy. AIUI, the cache for a given MH is unique per InternalFactory, but InternalFactory isn't widely shared (they are unique per binding IIRC). Is this improving cachability (and repeated injector constructions in tests, etc..) because we're removing the custom bytecode-generated Provider class? It seems like each distinct Injector will still have its own unique MethodHandles... but I guess that's more lightweight than a whole class definition per Provider? What value is the MH cacheability actually giving us if its unique per binding (or am I misunderstanding the scope of the caching)?

So you understand the caching correctly.  And indeed this will not address overhead in the case where a user constructs 2 identical or nearly identical injectors.  However within the scope of a single injector this caching is useful since we will call `getHandle` once for every injection point.

e.g.

```
@Provides Foo provideFoo(Bar bar) {}

@Provides OthjerFoo provideOtherFoo(Bar bar) {}
```

We will produce 2 `ProviderMethod` instances that each have a `SingleParameterInjector` pointed at the `InternalFactory` for `Bar`.  So by caching the `Bar` handle we ensure that each of those `ProviderMethod` instances can share the implementation handle.

So this is still valuable within the scope of an injector.  Not every binding is consumed multiple times, but some are and we don't want to recompute those.

So this really isn't improving repeated injector construction in tests.  It might help a little just since creating handles for each injector is a little cheaper, but it doesn't enable cross injector sharing.

The thing that does help the 'multiple injectors in a test' thing is dropping the bytecode generation since that was happening eagerly during injector creation (due to scopes), but now handle creation happens on demand.

Cross injector sharing is a compelling idea and it is something we do get with our fast-class code generation but not with MethodHandles.  Im really not sure how to actually achieve it.  We could create a global interning map for `InternalFactory` instances, but yikes the interning rules would be complex since `InternalFactory` instances are constructed in a complex way, can depend on arbitrary Injector state (and sometimes even the Injector itself).  So i think to do it would require some non-trivial effort and I am relatively inclined to think we shouldn't try (if a user has a problem they should switch to using childI injectors to share factories).

thoughts?

lukesandberg:
> > (Also, I ran through the depot last night, but forgot to flip the "enable this" flag to true... so running again now. Will see what it shakes out.)
> 
> depot is fairly happy. Some things that surfaced:
> 
> * Behavior changed for Provider instances that throw Errors (and possibly other things that throw errors?).  e.g, something like `bind(Foo.class).toProvider( { throw new AssertionError() } )`.. previously `injector.getInstance(Foo.class)` would surface that `AssertionError`, whereas now its wrapped inside a `ProvisionException`.  I'd argue that the `ProvisionException` is actually _more correct_, but we should avoid the behavior change for now and do it intentionally later if desired.  (This is commonly surfaced from @provides methods that threw Errors, but sometimes also just raw Provider instances. Presumably it'd be the same with a constructor that threw this way w/o any custom provider instance?)

Ah rats, forgot about this one.  Will update with a test and a fix.  I agree this seems odd


> * A very specific allocation leakage test which _should_ have shaken out "warmup" allocations, but now there appears to be new long-lived allocations per usage (despite a long-lived injector). I don't _think_ this PR is the cause (because AFAICT the "warmup" phase should pre-create all the requisite MHs), but will need to dig in further.  Not blocking this PR, but maybe you see something in this PR that's always creating new data per injection.

Interesting.  I wonder if it is the allocations related to the lazy MethodHandles construction?  That only happens when they actually call `get`.  There are also things like the `Intitializable` and `Singleton` bindings that will re-bind themselves after succesfull provisioning.  Beyond that (which honestly the warmup should take care of...), im not sure what it would be.

core/src/com/google/inject/spi/BindingSourceRestriction.java
lukesandberg: ok, i was confused before. this appears to be an incompatibiltiy with `ecj` which is used by the vscode java plugins... sigh.  It looks like there are number of issues where ecj reports and error and javac does not (and visa versa).

Added a comment to that effect.  I can revert if you prefer

sameb (approved):
(marking the latest iteration approved again, to be able to pull/test it.)

core/test/com/google/inject/ProvisionExceptionsTest.java
```diff
@@ -152,20 +170,70 @@ Exploder exploder() throws IOException {
...
+      throw new OutOfMemoryError("uh oh");
     }
   }
 
+  // Demonstrate that different bindings do and do not wrap Errors thrown.
+  // Constructor and method injection does since the reflection API does it for us via
+  // InvocationTargetException.
+  // Provider instances do not since we have our own catch clauses and only catch RuntimeException
+  @Test
+  public void testErrorPropagation() {
```
sameb, line 205: The exception management changes fixed most issues, but there's still one diff that's surfacing. The code follows a pattern of 

```
Provider<A> aProvider = { throw SomeError(); }
class BProvider implements Provider<B> {
  @Inject BProvider(A a) { .. }
}
// setup module binding A to aProvider instance
// setup module binding B to BProvider.class
// ask injection to get an instance of B
// expected SomeError, currently see ProvisionException
```

lukesandberg: Fun! This one was a little trickier but not too bad.  Basically we were placing the catch block around both the constructor binding and the constructor parameter bindings and it really needs to be just around the constructor invocation.

Found a similar issue with method injection

Finally found an inconsistency between constructor and provides injection in how exception sources are reported.  Documented that with a test.

lukesandberg: Oh also staring at this revealed an issue with the reflective fallbacks as well (we weren't always unwrapping InvocationTargetException), so that is fixed as well.

sameb (approved):
approving to pull internally again...

lukesandberg:
Just repushed to fix the javadoc issue flagged by bazel CI... sorry

i guess ill try to get the bazel build working locally again

sameb (approved):
(approving to pull to internal repo)

sameb:
> > * A very specific allocation leakage test which _should_ have shaken out "warmup" allocations, but now there appears to be new long-lived allocations per usage (despite a long-lived injector). I don't _think_ this PR is the cause (because AFAICT the "warmup" phase should pre-create all the requisite MHs), but will need to dig in further.  Not blocking this PR, but maybe you see something in this PR that's always creating new data per injection.
> 
> Interesting. I wonder if it is the allocations related to the lazy MethodHandles construction? That only happens when they actually call `get`. There are also things like the `Intitializable` and `Singleton` bindings that will re-bind themselves after succesfull provisioning. Beyond that (which honestly the warmup should take care of...), im not sure what it would be.

Re: this case, the path causing issues seems to follow the following chain of logic:

1) Per request, something injects the Injector.
2) Code then calls injector.getBinding(Key<Foo>(){})
3) Code then calls thatBinding.getProvider()
4) Code then calls thatProvider.get()

That *seems* to trigger `InternalMethodHandles.MethodHandleProvider.get()` to call `getHandle().invokeExact(currentContext)`, which leads the following reverse call-chain:

```
java.base/java.lang.invoke.LambdaForm.create(LambdaForm.java:342),
java.base/java.lang.invoke.LambdaForm.customize(LambdaForm.java:465),
java.base/java.lang.invoke.MethodHandle$1.apply(MethodHandle.java:1858),
java.base/java.lang.invoke.MethodHandle$1.apply(MethodHandle.java:1856),
java.base/java.lang.invoke.MethodHandle.updateForm(MethodHandle.java:1878),
java.base/java.lang.invoke.MethodHandle.customize(MethodHandle.java:1856),
java.base/java.lang.invoke.MethodHandle.maybeCustomize(MethodHandle.java:1846),
java.base/java.lang.invoke.Invokers.maybeCustomize(Invokers.java:633),
java.base/java.lang.invoke.Invokers.checkCustomized(Invokers.java:627), 
com.google.inject.internal.InternalMethodHandles$MethodHandleProvider.get(InternalMethodHandles.java:902)
```

(note the line # is from a prior snapshot, not the latest)

So either something in per-request Injector.getBinding().getProvider().get() is defeating MH caching, or the use of `invokeExact` is causing unique long-lived "LambdaForm"s to be created per usage.

(FYI I am probably going to submit this PR as-is since it's flag-guarded and is big enough for now, and we can follow up w/ fixes on another PR.  I'm also going on vacation starting tomorrow until the 18th, so will resume looking at this when I get back.)

lukesandberg:
interesting, ill see if i can write a test (in another PR!) to replicate this.

`invokeExact` is a somewhat magical method `@PolymorphicSignature` and performs custom linking.  As expected calling `invokeExact` on a non-constant (essentially `volatile`) `MethodHandle` is slower, and i think part of that is this 'handle customization' work.  From reading some old [threads](https://mail.openjdk.org/pipermail/mlvm-dev/2017-May/006755.html), i see that this 'customization' is triggered when a method is JIT'd that contains a call to a MethodHandle that is not visible to the jit.  This applies to `MethodHandleProvider` since that `get` method will get compiled but `getHandle` is always returning (from the JITs perspective) a new handle.

So the JVM avoids special handling, but tries a bit harder after 127 invocations (controlled by `java.lang.invoke.MethodHandle.CUSTOMIZE_THRESHOLD` which defaults to 127).  So I'm guessing that that is how we defeat the tests warmup.

So i _think_ the solution is one of....

1. Go back to bytecode generating the provider subtypes.  This gives the JIT a much easier job and can trivially 'customize' on first compile, but of course comes with all kinds of other performance and bootstrapping tradeoffs.
2. Hide the handles behind a 're-binding' callsite?
   - This would be some kind of `tableSwitch` to pick the delegate handle, but also using some kind of `WeakReference` to avoid pinning injectors... sigh.  Alternatively we could bytecode gen one MethodHandleProvider class per injector....?
   - This solution is cautiously recommended [here](https://mail.openjdk.org/pipermail/mlvm-dev/2018-February/006816.html) after a very long discussion.
   - However, for us, we would be re-binding this callsite like 1000's of times... so it would induce a TON of work in the JIT.
   - And afaict this is just re-implementing what MethodHandle 'customization' is meant to do anyway.
3. Somehow force 'customization' to happen sooner?  This is probably a bad tradeoff since a fair number of these `Provider` instances would only ever be invoked a small number of times.
4. Decide that the test in question is simply not important?  I'm guessing the test is using some kind of AllocationAgent that is sensitive to the allocations in Java code but not the JIT compiler... but that is more or less what they are measuring here.   So we could fix the test or just disable customization for them by setting `-Djava.lang.invoke.MethodHandle.CUSTOMIZE_THRESHOLD=-1`. which will give them more deterministic behavior but worse performance.

sameb:
> 1. Go back to bytecode generating the provider subtypes.  This gives the JIT a much easier job and can trivially 'customize' on first compile, but of course comes with all kinds of other performance and bootstrapping tradeoffs.

This would be sad. :-(

> 2. Hide the handles behind a 're-binding' callsite?
>    
>    * This would be some kind of `tableSwitch` to pick the delegate handle, but also using some kind of `WeakReference` to avoid pinning injectors... sigh.  Alternatively we could bytecode gen one MethodHandleProvider class per injector....?
>    * This solution is cautiously recommended [here](https://mail.openjdk.org/pipermail/mlvm-dev/2018-February/006816.html) after a very long discussion.
>    * However, for us, we would be re-binding this callsite like 1000's of times... so it would induce a TON of work in the JIT.
>    * And afaict this is just re-implementing what MethodHandle 'customization' is meant to do anyway.

Seems unfortunate but maybe necessary?

> 3. Somehow force 'customization' to happen sooner?  This is probably a bad tradeoff since a fair number of these `Provider` instances would only ever be invoked a small number of times.

+1 bad tradeoff.

> 4. Decide that the test in question is simply not important?  I'm guessing the test is using some kind of AllocationAgent that is sensitive to the allocations in Java code but not the JIT compiler... but that is more or less what they are measuring here.   So we could fix the test or just disable customization for them by setting `-Djava.lang.invoke.MethodHandle.CUSTOMIZE_THRESHOLD=-1`. which will give them more deterministic behavior but worse performance.

I think effectively suppressing the test (whether ignoring LambdaForms or reducing the threshold & expecting an exact # of things) depends on the consequences in real-life behaviors.  We need to ensure that the long-lived LambdaForms created per-request (from the invokeExacts) aren't created *per every request in a unique way per request*.  Maybe the problem is just that the warmup threshold is lower than the customize threshold, so its being inlined *after the warmup*, but we're not actually creating a new LambdaForm *per request*, we're just doing it "per unique signature used in practice" (which shouldn't really grow, IIUC).  WDYT?

Issue #1179, opened 2018-04-21
Guice persist JpaLocalTxnInterceptor gets stuck when connection to database drops during transaction


bgooren:
We have an interesting case in some production code: inside a method annotated with transactional some operation takes much longer than expected, and our database connection pool closes the database connection.
A query inside the method then throws an exception ("Connection has already been closed").

The JpaLocalTxnInterceptor takes care of starting a unit of work (we don't manage it ourselves in this instance). I would expect the unit of work to have ended after the method call.

However, this is what happens:

- rollbackIfNecessary is called, which tries to execute a transaction rollback
- the rollback throws an exception ("Connection has already been closed")
- as the transaction is still seen as active, the unit of work is not closed (line 82 in JpaLocalTxnInterceptor)
- every next call to the annotated method tries to reuse the unit of work and fails

The above can be circumvented by managing the unit of work ourselves. But that means that we cannot rely on the "it just works"-behavior of the transactional annotation, which is a shame.
It also means more boilerplate code in every location where we now rely on the transactional annotation to do the work for us.

Ideally the interceptor should end the unit of work, even when it cannot perform a rollback.

Vogel612:
#1136 seems somewhat related. The fix itself is similar

bgooren:
Somewhat, yes. In my case neither a commit() or rollback() will work, as the underlying connection was closed. This is not handled by JpaLocalTxnInterceptor, which means it gets stuck reusing a broken connection. In your case you want to prevent a call to commit() after your own call to rollback(), to prevent an exception from being thrown.

cs-devil:
@bgooren have you find any way to solve this issue.

Vogel612:
@cs-devil there is an open PR from yours truly to fix the issue, which has been languishing for a while awaiting a restart of an intermittently failing CI [here](https://github.com/google/guice/pull/1137)

Qwertee:
Is there any fix for this that is planned to be merged? We have run into the exact same problem with connection closed errors which makes it impossible to use Guice's `@Transactional` annotation since dead `EntityManger`'s get locked into our threads.

To get around this, we built our own method interceptor which manages the `UnitOfWork` lifecycle in a more appropriate way for our use-case.

bgooren:
@Qwertee not that I'm aware of.

[...]

PR #1917, opened 2025-09-09
Fix EntityManager leak in ThreadLocal after failed transaction

### LINKED ISSUE #1916: EntityManager leak in ThreadLocal when connection is lost during transaction
## Summary
When using guice-persist with a web server such as Jetty, a broken EntityManager may leak into subsequent requests due to flawed cleanup logic in the transaction interceptor.

## Bug description
- A request is handled by a Jetty thread.
- Guice interceptor opens a database connection and stores the EntityManager in a ThreadLocal.
- A network blip occurs, breaking the database connection mid-transaction.
- On commit/rollback, an exception is thrown.
- The interceptor cleanup logic checks if (transaction.isActive()).
- Because the connection is already dead, this returns false.
- As a result, the cleanup logic skips removal of the EntityManager from the ThreadLocal.
- The thread is reused by Jetty for a new request.
- The broken EntityManager is still present in the ThreadLocal.
- The new request fails immediately.

## Expected Behavior
If the connection is no longer open, the EntityManager should be removed from the ThreadLocal during cleanup.

## Actual Behavior
If the transaction is no longer "active" (e.g., due to connection loss), cleanup is skipped and the broken EntityManager persists in the ThreadLocal, poisoning subsequent requests.

## Impact
- Causes intermittent request failures.
- Leads to hard-to-debug production issues when connections are lost.
- Especially problematic in long-lived thread pools like Jetty, Tomcat, or Netty.

## Similar issue
https://github.com/google/guice/issues/1179

## Proposed Fix
Remove the EntityManager from the thread’s ThreadLocal in the cleanup logic if the connection is no longer open

## Contribution
I plan to open a PR with a fix for this issue shortly. Feedback on the proposed approach is welcome before I finalize the patch.

SantanuKar43:
This PR fixes a bug in JpaLocalTxnInterceptor where a broken EntityManager could remain in a thread’s ThreadLocal after a failed transaction, leading to subsequent requests failing when the same thread is reused by the web server (e.g., Jetty).

## Problem
- Guice Persist stores the EntityManager in a ThreadLocal.
- If a transaction fails due to a database connection issue (e.g., network blip):
- txn.isActive() incorrectly returns false because the connection is already dead.
- The cleanup logic is skipped due to this check.
- The broken EntityManager is never removed from the ThreadLocal.
- When the thread is reused for another request, the stale EntityManager is returned, causing immediate failures.

## Fix
- Updated cleanup logic to always remove the EntityManager from the ThreadLocal if this interceptor started the work, regardless of txn.isActive().
- Reorganized the try/catch/finally blocks so that commit/rollback semantics remain correct while ensuring cleanup runs after transaction handling.
- Ensures no broken EntityManager instances leak into future requests.

## Related Issue

Fixes https://github.com/google/guice/issues/1916 and https://github.com/google/guice/issues/1179

### DIFF
--- a/extensions/persist/src/com/google/inject/persist/jpa/JpaLocalTxnInterceptor.java
+++ b/extensions/persist/src/com/google/inject/persist/jpa/JpaLocalTxnInterceptor.java
@@ -65,6 +65,12 @@ class JpaLocalTxnInterceptor implements MethodInterceptor {
     try {
       result = methodInvocation.proceed();
 
+      if (txn.getRollbackOnly()) {
+        txn.rollback();
+      } else {
+        txn.commit();
+      }
+      return result;
     } catch (Exception e) {
       // commit transaction only if rollback didnt occur
       if (rollbackIfNecessary(transactional, e, txn)) {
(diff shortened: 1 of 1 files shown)


SantanuKar43:
@sameb @lukesandberg can you please review this?

SantanuKar43:
@cushon can you please review this or help assign to someone who can?

